PR #3, opened 2017-07-27
PLT-6985: Add Elasticsearch to Mattermost Kubernetes.

grundleborg:
Add Elasticsearch to Mattermost Kubernetes.

https://mattermost.atlassian.net/browse/PLT-6985

This more-or-less works, but has definite room for improvement.

Major Known Issues:
* Elasticsearch does not use a Persistent Volume Claim, so no data is
  stored persistently.
* If the Elasticsearch pods come up after the App Server pods, the app
  server has already failed to connect to them, so assumes Elasticsearch
  is disabled. Workaround is to kill the App Server pods so they get
  restarted. Proper fix would be some kind of probe that checks they are
  up similar to the one for the database.

### DIFF
--- a/mattermost-helm/charts/mattermost-app/templates/config.json
+++ b/mattermost-helm/charts/mattermost-app/templates/config.json
@@ -48,12 +48,12 @@
         "EnableUserStatuses": true,
         "ClusterLogTimeoutMilliseconds": 2000
     },
-    "ElasticSearchSettings": {
-        "ConnectionUrl": "",
-        "Username": "elastic",
-        "Password": "changeme",
-        "EnableIndexing": false,
-        "EnableSearching": false,
+    "ElasticsearchSettings": {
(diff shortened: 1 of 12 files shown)


grundleborg:
@crspeller @coreyhulen tagging you here since I don't have any permissions on this repo (can't even request reviews).

coreyhulen:
Cool, @grundleborg you should have access now.

grundleborg:
Pre-merge ToDo List:
1. Use an init container to delay launching the app containers until the elasticsearch cluster is ready.
2. Config value to make using Elasticsearch optional when deploying the chart.

grundleborg:
Fixed both of the above items. This is now ready to merge, once you guys are happy with it @coreyhulen @crspeller

grundleborg:
Bump. How do you guys feel about merging this soon?

PR #5, opened 2017-08-07
PLT-7169 Read replicas for mysql

escardin:
This changes mattermost-kubernetes to have read replicas for mysql. It's a copy of a pull request for a helm incubator repo (https://github.com/kubernetes/charts/pull/766). Which in turn is based on https://kubernetes.io/docs/tasks/run-application/run-replicated-stateful-application/

I removed mattermost-db in favour of this, as once the PR is merged into helm, this code can be removed, reducing maintenance problems. This is the reason it's called mysqlha.

I modified how the values for the read replica are passed in so that the chart is able to be used as it is in the pr without modification (pending a pr for that chart https://github.com/jpoon/charts/pull/3). values are set in the mysqlha section of the root values.yaml, and will be hoisted out into the global section via a section in requirements.yaml. This is a little more portable, and provides a way to use the password generation in the mysql chart to have a more secure secret.

replica count is controlled by mysqlha.mysqlha.replicaCount, which defaults to 3 (this includes the mysql master)

This change also includes using persistent volumes, which are also configurable.

Should resolve https://github.com/mattermost/platform/issues/6959


crspeller:
So this new chart is not in the official repository yet. But we should be able to grab it when it is by using `helm dependency update` right?

escardin:
Not quite. Right now there is a copy in the repo that is identical to the outstanding PR. When it merges, you can delete the code in this repo, add the incubator dependency, then do `helm dependency update`. For now you don't need to do any dependency updates.

I looked into hosting the dependency in such a way that you would only have to change what repository the chart is coming from, but it involved some things I felt were more problematic than just keeping the code here.

crspeller:
Oh alright. I misread, thought it was to make it official. Strange that kubernetes doesn't already support RRs for mysql.

coreyhulen:
@crspeller can we run a load test with this PR to see if it's good?  I guess we might need to check the fork or locally? not sure how that works with the build.

crspeller:
@escardin Can you do a rebase?

escardin:
done.

ccbrown:
Closing as this functionality has been resubmitted in https://github.com/mattermost/mattermost-kubernetes/pull/10.

PR #10, opened 2017-11-28
PLT-7169: MySQL HA

grundleborg:
Updated MySQL HA PR using the kubernetes/charts incubator chart.

This still doesn't quite work due to the nasty interaction between the mattermost-app chart and the statefulset that I mentioned on pre-release. Opening the PR in case anyone else wants to investigate while I'm away.


ccbrown:
Commandeering this: https://github.com/mattermost/mattermost-kubernetes/pull/13.

PR #13, opened 2018-02-05
ABC-37: MySQL Read Replicas

ccbrown:
Rebase of @grundleborg's PR that I'll be trying to finish up soon:

> The newer PR (the one submitted by me) is the one to start from. It is based off the older one, but uses the MySQL-HA chart from the Kubernetes Incubator instead of copying it across into our repo as in the older PR.
> 
> It very nearly works. In fact, it works flawlessly in my testing as long as the Mattermost server doesn't try to connect to the MySQL database while the stateful set is still spinning up. However, something goes wrong that breaks the stateful set launch process once the Mattermost server tries to connect to it, so often it gets stuck with only 0 or 1 working read-replicas instead of the 2 that are configured by default.
> 
> I don't know what the issue is - I had a bit of a look into it, but then had to move to other priorities. I would suggest that the place to start is with my PR, and then to investigate why the stateful set startup is interacting badly with the Mattermost Server. It might need fixing upstream on the MySQL-HA chart, or it might be something in our charts or my PR - I don't really know.

– @grundleborg


cpanato:
closing in favor of https://github.com/mattermost/mattermost-kubernetes/pull/25

thanks @ccbrown  🎉

PR #14, opened 2018-03-05
Update values.yaml

daanlevi:
Without the key you get an error deploying with helm
Error: render error in "mattermost-helm/charts/mattermost-elasticsearch/templates/pvc.yaml": template: mattermost-helm/charts/mattermost-elasticsearch/templates/pvc.yaml:1:14: executing "mattermost-helm/charts/mattermost-elasticsearch/templates/pvc.yaml" at <.Values.persistence....>: can't evaluate field enabled in type interface {}


coreyhulen:
Thanks, sorry, missed this.

Eventually we would like them to be in the official repo, but for now they are used more for load testing and are a work in progress.

PR #15, opened 2018-04-14
Adding persistence section in mattermost-elastic values

jespino:
If you don't have it defined the mattermost-kubernetes fails on `helm install
mattermost-helm`.


jespino:
Duplicated, sorry.

PR #18, opened 2018-04-26
Cleaning up .gitignore

jespino:
There was a lot of things without any relation with the mattermost-kubernetes
project, generating some weird cases like config.json file was in the list of
ignored files for this repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,62 +1,9 @@
-
-config.json
-config.yaml
-db-password.txt
-db-root-password.txt
 mattermost.mattermost-license
 *.mattermost-license
 
-logs
+dist/
+
 .DS_Store
-node_modules
-/dist
-/webapp/dist
-npm-debug.log
-
-web/static/js/bundle*.js
-web/static/js/bundle*.js.map
-web/static/js/libs*.js
-
-config/active.dat
-config/config.json
-
-# Enteprise imports file
-imports.go
-cmd/mattermost/imports.go
-
-# Build Targets
-.prebuild
-.npminstall
-
-# Compiled Object files, Static and Dynamic libs (Shared Objects)
-*.o
-*.a
-*.so
-
-# Folders
-_obj
-_test
-
-# Architecture specific extensions/prefixes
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,62 +1,9 @@
...
-*.exe
-*.test
-*.prof
-
-# Log files
-*mattermost.log
-*npm-debug.log*
-
-.tmp
+.DS_Store
```
lieut-data: IMO, the only valid entry in this file is `dist/` (unless we generate the license files?)

I don' think we should `.gitignore` files that aren't part of the repository. They are better handled by a developer/user's `.gitignore_global` or some such.

jwilander: Then won't we be relying on every single contributor to correctly set up their own `.gitignore_global`? I get that it would be ideal but if it's introducing another barrier to contributing I think it would be much better to just add the basic stuff to our .gitignore

jespino: I agree with @jwilander , I prefer to have a gitignore in the repo.

lieut-data: That's an angle I hadn't considered. Has that been a problem in the past?

On one hand, I'm not particularly convinced it's a high barrier to entry. I'd be more inclined to take advantage of the "teachable" moment and make that developer more effective going forward, even if that just means shifting them away from a blind `git add --all`. Plus, I note that we're missing some pretty basic stuff already: even webapp doesn't have `.vscode` ignored.

On the other hand, I'm not hung up on this. If we broke out our `.gitignore` files into "files we generate but ignore" (useful documentation) and a common "other files" section that we (try to) keep in sync across our repositories as we encounter new such files, it wouldn't be a big deal.

lieut-data: Looking at the (recently) canonical example of an open source project, it looks like my opinions are stale on this front: https://github.com/kubernetes/kubernetes/blob/master/.gitignore contains lots of such files.

I won't hold up the review, though it would be good to converge on making this file "exhaustive" (for some such definition) vs. somewhat random.

PR #19, opened 2018-04-26
Fix elasticsearch chart

jwilander:
@crspeller I know we originally wanted to include elasticsearch via requirements.yaml but it's not in any helm repository already and we've already diverged a decent amount from the master of the helm chart this was based off. The main difference being that helm chart is using elasticsearch major version 6 and we're on major version 5.

I was also going to make all these changes in our fork of that original helm chart but I realized I don't think that's worth all the extra overhead and effort if we're just maintaining this chart separately anyways.

Eventually I think it would be nice if someone start maintaining a helm repo that we could include but for now I think it's easier if we just keep this in here.


crspeller:
Agreed. If we can't reasonably import it, I think this is fine.

PR #24, opened 2018-05-07
MM-10213 Replace with mattermost-proxy chart with nginx-ingress, add kube-lgeo for SSL certs

jwilander:
* Replaced mattermost-proxy with the nginx-ingress stable chart
* Added some configuration info so I wouldn't forget it
* Moved jobserver and notifications configuration under `global.features` because the loadtest chart was complaining about not having access to it
* Added kube-lego chart for generating TSL certs. FYI kube-lego is deprecated but the new one, cert-manager, won't work as a helm dependency due to the order helm installs charts. So until that's fixed kube-lego is are best option


jwilander:
Just realized I butchered that commit message :sweat_smile:

cpanato:
I will merge it to get the update repo :)

PR #40, opened 2018-06-27
Deploy MM Chart and apply basic tests and teardown

cpanato:
fixes: https://mattermost.atlassian.net/browse/MM-11057

### DIFF
--- a/mattermost-helm/values.yaml
+++ b/mattermost-helm/values.yaml
@@ -70,10 +70,6 @@ mysqlha:
     enabled: false
     size: 10Gi
 
-tags:
-  core: true
-  metrics: false
-
 mattermost-app:
   replicaCount: 2
   image:
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,15 +1,110 @@
 #!/usr/bin/env groovy
 
 def gitCommit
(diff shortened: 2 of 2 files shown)


mattermost-helm/values.yaml
```diff
@@ -171,7 +167,7 @@ minio:
   secretKey: mattermostadmin
   nodeSelector: {}
   persistence:
-    enabled: true
+    enabled: false
```
crspeller: We should have reasonable defaults in the `values.yaml` not defaults setup for our build server. We should modify this just for the test.

cpanato: Ok, but in my opinion, all persistence should set to false. 2/5. I will revert that and set in the test for now

cpanato:
@jwilander PTAL

mattermost-helm/values.yaml
crspeller: Ok. But this PR is about setting up the build, not changing defaults, so you can see why I thought it was related.

crspeller:
@cpanato Doesn't seem to be building.

cpanato:
was missing one thing :(. now it is fixed

PR #41, opened 2018-07-01
MM-8694: Add PostgreSQL support

jespino:
Add PostgreSQL support, and allow to configure it throug values. This would be the first step for other options like patroni backend.

To test it, is easy, you simply create a values.yaml file with the content:

```
global:
  dbBackend: "postgresql"

mysqlha:
  enabled: false

postgresql:
  enabled: true
```

and then run:

```
make package
helm install dist/mattermost-helm-0.2.4.tgz -f values.yaml
```

This is for a single postgresql server.

### DIFF
--- a/mattermost-helm/charts/mattermost-app/templates/config.json
+++ b/mattermost-helm/charts/mattermost-app/templates/config.json
@@ -84,9 +84,15 @@
         "MaxNotificationsPerChannel": 1000
     },
     "SqlSettings": {
+        {{- if eq .Values.global.dbBackend "mysqlha" -}}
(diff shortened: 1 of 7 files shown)


cpanato:
looks good, i will review later this week

jwilander:
Is there any existing postgres charts that have replication?

jespino:
@jwilander Yes, that will be my next step. In incubator exists a chart call "patroni" which is the Postgresql cluster from Zalando. My plan is to add that soon.

jespino:
No longer needed, because the decision has been to support only one database included directly in the chart

PR #42, opened 2018-07-04
MM-10846 Make link previews configurable

jwilander:


mattermost-helm/values.yaml
```diff
@@ -29,6 +29,8 @@ global:
         # Hosted push notification service. Requires an enterprise license.
         # More details at https://docs.mattermost.com/mobile/mobile-hpns.html.
         useHPNS: false
+    linkPreviews:
+      enabled: false
```
cpanato: maybe just add `linkPreviews: false` instead of having another level

jwilander: That's what we do with all the others, I'd rather be consistent: https://github.com/mattermost/mattermost-kubernetes/blob/6410af8588d4692f021211f3ce9686b46890a5ac/mattermost-helm/values.yaml#L12

cpanato: hummm, ok
but we should rewrite thos whole section and add some config level, like the mm-te. but lets do that in a followup

jwilander: Agreed, the whole thing needs a rewrite

PR #45, opened 2018-07-09
Remove mm-app and loadtest from charts folder and move to template folder

cpanato:
Making the chart more similar to others in the K8s. the main app should stay in the template folder with your addons (like loadtest / jobserver) others like mysql/elastic search if is not mandatory we should import as dependencies.

next step: refactor the values.yaml


cpanato:
@jwilander sure no rush , just let me know or merge this when it is ready

jwilander:
@cpanato when you get back: if you can rebase I can then go update ltops so that it works with these changes and we can get this merged

jwilander:
FYI I've started updating ltops to match these changes, should be done sometime today or tomorrow

PR #48, opened 2018-08-13
remove nginx-ingress and kube-lego

cpanato:
This is part of the k8s cluster configuration and the user can select
other ingress.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,15 @@ Mattermost Kubernetes (Beta)
 
 This is the helm chart for Mattermost Enterprise Edition. It is in beta and subject to changes but is used by Mattermost internally to run some CI servers.
 
+# Pre requisites
+
+To use the Mattermost Enterprise Helm Chart you will need first a running Kubernetes cluster or Minikube.
+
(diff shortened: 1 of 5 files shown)


pgold30:
Tried this, the ingress and kube lego are being generated anyway

cpanato:
@pgold30 did you run `helm dependency update` and then delete the .tar.gz files in the charts folder?

pgold30:
Yes, and also try from scratch, i was able to remove nginx and kube-lego installation settings ingress values to false, it seems that for some reason when ingress is on true nginx and kube-lego are deployed

cpanato:
@pgold30 I retest locally and it did not install the ngnix-ingress and kubelego using the default config

cpanato:
will add 😄

cpanato:
@jwilander @crspeller update the readme

cpanato:
@crspeller @jwilander ptal

jwilander:
@cpanato this means that ltops w/ k8s will also need `nginx-ingress` as a prerequisite. Maybe we should add to the ltops tool some code to check if it's installed already and if it's not, have it install `nginx-ingress`

cpanato:
@jwilander I dont think the ltops need the nginx ingress, because all the load we make is intra cluster usinf the cluster dns and not required an external endpoint.

jwilander:
We do want to be able to access the cluster externally if needed though

jwilander:
Chatted with @cpanato, right now the load test pods are pointed directly at the app servers so they don't need the ingress. We're going to create a ticket to point them at the ingress (for image caching, and because that's how real set ups will work) and create nginx ingress with ltops if needed

PR #51, opened 2018-08-30
[MM-11890] Replace Deployment to Statefull sets in order to use better the plugins

cpanato:
- replace the Deployment of the App to use the Stateful sets, then we can have volumes for the plugins and config and when scale the app it will not break

 - Make a tricky in the config.json. It copy the configmap when it is the first time, if restarts it checks if the configmap have the same sha256 as the config in the pod if have the same that means there is not change in the confimap and then still use the config in the pod which might have some user changes, like the plugins enable...
 
- enable the job server to run. (found some issues here, need some clarification)


mattermost-helm/templates/config.tpl
```diff
@@ -309,14 +309,22 @@
...
+    },
     "DataRetentionSettings": {
         "Enable": false
     }
     {{- if .Values.global.features.jobserver.enabled -}}
     ,
     "JobSettings": {
-        "RunJobs": false,
-        "RunScheduler": false
+        "RunJobs": true,
```
crspeller: There is a separate job server so shouldn't these be disabled?

cpanato: ahhh got it

PR #54, opened 2018-09-14
Load test point to ingress and other updates

cpanato:
jira: https://mattermost.atlassian.net/browse/MM-11906

### DIFF
--- a/mattermost-helm/values.yaml
+++ b/mattermost-helm/values.yaml
@@ -1,18 +1,33 @@
 global:
-  # Set siteUrl to the URL your users will use to access Mattermost
-  siteUrl: ""
-  filesAccessKey: "mattermostadmin"
-  filesSecretKey: "mattermostadmin"
+  siteUrl: "http://mattermost.example.com"
+  siteName: "Mattermost"
   mattermostLicense: "THIS STRING SHOULD BE REPLACED WITH THE CONTENTS OF YOUR LICENSE FILE"
+
+  # Settings for general configuration of MM
+  # For more details: https://docs.mattermost.com/administration/config-settings.html
+  enableTeamCreation: true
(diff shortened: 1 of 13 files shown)


mattermost-helm/values.yaml
```diff
@@ -1,18 +1,8 @@
 global:
-  # Set siteUrl to the URL your users will use to access Mattermost
-  siteUrl: ""
-  filesAccessKey: "mattermostadmin"
-  filesSecretKey: "mattermostadmin"
   mattermostLicense: "THIS STRING SHOULD BE REPLACED WITH THE CONTENTS OF YOUR LICENSE FILE"
   features:
```
crspeller: Why move some of this stuff away from here? It seems like it's just spreading it around the config file making it harder to find instead of having it all in one place here at the top.

cpanato: - `filesAccessKey` and `filesSecretKey` was not use at all
- the `siteUrl` is only used in the config.json

global section has a lot of things it is hard to read and navigate and also most of the things there are not "global". I'm trying just to put the things in a order and in the right place.

what is your suggestion? keep the siteUrl in the global?

crspeller: Just wanting to simplify the config. Your PR spreads the usefull configuration settings all over the place. I get the scoping stuff, but wondering if it's worth it to just bring a bunch of useful globals to the top. Gitlab seems to be doing that with their chart: https://gitlab.com/charts/gitlab/blob/master/values.yaml

cpanato: right, will revert

cpanato: did the change @crspeller ptal and let me know what you think.

PR #57, opened 2018-11-12
Add extra initContainer to download and install the plugins

cpanato:


mattermost-helm/values.yaml
```diff
@@ -153,6 +153,33 @@ mattermostApp:
...
+    #         cd /mattermost/plugins
+    #         tar -xzvf ${plugin_tar}
+    #         rm -f ${plugin_tar}
+    #       done
+    #   volumeMounts:
+    #   # Need to match the volume mount for plugins volumes
+    #   - name: mattermost-plugins
+    #     mountPath: /mattermost/plugins/
+    #   - name: mattermost-plugins-client
+    #     mountPath: /mattermost/client/plugins/
```
jwilander: Is this all commented out on purpose?

cpanato: yes, because you don't know which plugins the user will download.

PR #61, opened 2019-02-20
MM-13091 Review, improve and test the push proxy docker image

stylianosrigas:
> MM-13091 Review, improve and test the push proxy docker image
 
- Current Dockerfile follows best practises
- Ιmpovements made include:
  - Change of PP_VERSION variable to APP_VERSION 
  - Combining of all Docker Volume creation entries in a single entry, 
  - Move of ENTRYPOINT entry to the end of the Dockerfile to follow Dockerfile best practises.
- The docker image was recreated and tested to make sure the server works as expected.

Fixes https://github.com/mattermost/mattermost-server/issues/10314

### DIFF
--- a/docker/push-proxy/Dockerfile
+++ b/docker/push-proxy/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 ENV APP_VERSION=4.4.0
 ENV PATH="/mattermost-push-proxy/bin:${PATH}"
--- a/docker/push-proxy/Dockerfile
+++ b/docker/push-proxy/Dockerfile
@@ -1,6 +1,6 @@
 FROM ubuntu:16.04
 
-ENV PP_VERSION=4.4.0
+ENV APP_VERSION=4.4.0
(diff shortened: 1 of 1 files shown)


docker/push-proxy/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM ubuntu:16.04
```
cpanato: can we use another image like `alpine`?

@jwilander do you know if exist any external dependencies?

jwilander: None that I'm aware of

stylianosrigas: @cpanato @jwilander I will test with alpine and update ;)

stylianosrigas: I have tested with Alpine and it was causing issues to the existing push-proxy application. Therefore, I moved to using version 18.04 of Ubuntu, which is the new Ubuntu Minimal version and more details can be found here -> https://blog.ubuntu.com/2018/07/09/minimal-ubuntu-released. The total size of this image is 29MB.

PR #62, opened 2019-02-22
MM-14263 Review, improve and split out push proxy helm chart

stylianosrigas:
> https://github.com/mattermost/mattermost-server/issues/10313

- Moving the push-proxy chart out of the enterprise-edition directory.
- Refactoring code to support push-proxy independent deployment. 
- Update of README to cover new changes.
- The push-proxy helm chart should be added in the requirements.yaml of mattermost-enterprise-edition when the chart will move in a central location. Combination of local and remote requirements is not supported at the current helm version.


charts/mattermost-enterprise-edition/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Mattermost Enterprise server with high availitibity.
 name: mattermost-enterprise-edition
-version: 0.7.1
+version: 0.7.2
```
cpanato: it is a relative medium change would bump the minor 
```suggestion
version: 0.8.0
```

stylianosrigas:
@cpanato I have updated the Chart version for mattermost-enterprise-edition. Furthermore, I have updated all mattermost-push-proxy templates to use the new label format, which is defined in the helpers file.

stylianosrigas:
@cpanato Tests added for Mattermost-push-proxy

charts/mattermost-push-proxy/templates/configmap-apple-cert.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "mattermost-push-proxy.fullname" . }}-apple-cert-path
+  labels:
+    chart: {{ include "mattermost-push-proxy.chart" . }}
```
cpanato: can we use the new labels? like https://github.com/mattermost/mattermost-kubernetes/blob/master/charts/mattermost-enterprise-edition/templates/deployment-mattermost-app.yaml#L6

stylianosrigas: @cpanato Mattermost-push-proxy adapted to use new labels. Let me know if anything is missing.

PR #64, opened 2019-03-05
udpate e2e test to use working local storage

cpanato:
- Update the kind to use the newer version of the hostpath provisioner, so we can test the chart creating volumes
- fix the initContainer

### DIFF
--- a/tests/e2e-kind.sh
+++ b/tests/e2e-kind.sh
@@ -74,18 +74,15 @@ install_tiller() {
     echo
 }
 
-install_hostpath-provisioner() {
-    # kind doesn't support Dynamic PVC provisioning yet, this is one way to get it working
-    # https://github.com/rimusz/charts/tree/master/stable/hostpath-provisioner
+install_local-path-provisioner() {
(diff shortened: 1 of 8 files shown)


cpanato:
path provisioner is: `Local Path Provisioner provides a way for the Kubernetes users to utilize the local storage in each node.`
We need to set up some storage in the k8s cluster, in the hosted solutions it provides a default one, but in a clean k8s this might not be available and you will need to set up one

tests/e2e-kind.sh
```diff
@@ -74,18 +74,15 @@ install_tiller() {
     echo
 }
 
-install_hostpath-provisioner() {
-    # kind doesn't support Dynamic PVC provisioning yet, this is one way to get it working
-    # https://github.com/rimusz/charts/tree/master/stable/hostpath-provisioner
+install_local-path-provisioner() {
+    # kind doesn't support Dynamic PVC provisioning yet, this is one ways to get it working
```
lieut-data: The original `way` was correct, grammatically.

cpanato: 😄

PR #67, opened 2019-03-28
Remove minikube from install instructions

jwilander:
It's not very helpful to have minikube included and just makes the setup instructions seem more confusing.


jwilander:
I think just linking out to the official docs for kind/minikube/whatever is good enough

cpanato:
Sounds good

PR #106, opened 2019-08-25
MM-17927: Move to circleci. Use different iproute package.

metanerd:
https://github.com/mattermost/mattermost-server/issues/11922
Encountered docker build failing for https://github.com/mattermost/mattermost-load-test. It apt couldn't find iproute package, instead it suggested using iproute2. Then docker build succeeded.


cpanato:
~I think we dont use this docker image anymore, only the one in the load-test repo.~
i see. I think we should use the dockerfile here https://github.com/mattermost/mattermost-load-test/blob/master/Dockerfile
maybe replace the file in the load-test with this one.

Also this will be better because the pipeline dont need to clone another repo

@jwilander any thoughts?

jwilander:
Simplifying the pipeline always makes sense to me 👍 what do you think @metanerd ?

cpanato:
then we remove the docker file from this repo and replace in the other repo

metanerd:
I am not sure this wouldn't destroy user space. On the other hand it is also not clear to me what the benefits are from the non bundled alpine Dockerfile. 
I cannot find anything in the git commit message, why it was added.

PR #108, opened 2019-08-29
add PVC and volume mount for plugins directory

agusl88:
Adds persistency to the plugins folder (/mattermost/plugins)

I'm not sure about the volume size defined on the values.yaml, i think plugins are mostly small golang scripts so 1gb should be enought, but i don't have much experience with them since i just starting using mattermost.

Do we also need persistency to the ClientDirectory path? (/mattermost/clients/plugins)


onpaws:
Sorry for the zombie revive, wanted to share the solution I just landed on in case anyone else gets `mkdir client/plugins/com.mattermost.calls: permission denied`:

image

Turns out the permissions for the plugins PVC default to `root` instead of `mattermost`, which breaks things.

Presumably an `initContainer` would offer a more hands-free fix, but what worked for me as a quick fix was to temporarily attach the volume to a new pod + `chmod -R 2000:2000 /path/to/vol`. Reattach back to Mattermost, and after init the plugins work as they should 🎉 

image

Hope it helps someone

liondy:
hello @onpaws I'm facing the same issue now (I need Mattermost Board to run) but failed to mkdir focalboard at client/plugins. Can you help me a solution for this? Thankyou

I have tried to add extraInitContainers (from this link https://github.com/mattermost/mattermost-helm/issues/120#issuecomment-549968396), and change the mountPath to client/plugins, but the volumeMounts name not found. Or is my workaround approach is wrong? Thankyou

onpaws:
Hi, it's been a while since I worked on this.
> failed to mkdir focalboard at client/plugins

It sounds like you have a permissions issue, but without the error its hard to say.
I would still suggest what I said in my comment:
> temporarily attach the volume to a new pod + chmod -R 2000:2000 /path/to/vol

Probably some ubuntu base image or similar is good enough.

PR #110, opened 2019-09-06
fix the mattermostApp externalPort variable

unlikelygeek:
helm install fails when service is set to loadBalancer


jasonblais:
@unlikelygeek Thanks for signing the CLA! Let us know if you have any questions about bumping the helm chart version, we'd be happy to help. 

Really appreciate your contribution!

PR #132, opened 2020-01-17
MM-20705 Use correct apiVersion based on K8s version in mattermost-team-edition

khos2ow:
#### Summary

Originally in #126 there was an attempt to change the chart to consider deprecated Kubernetes API (namely in 1.16) this PR is to complete that. Instead of completely drop the support for older API version this PR reads the Kubernetes Version from Helm's `Capabilities` built-in objects and executes `semverCompare` against the underlying K8s version and known version in which given API was deprecated.

~This also supports both Helm v3 and v2 (as the structure of `Capabilities` in v3 was slightly changed)~ Helm v3 support was dropped (https://github.com/mattermost/mattermost-helm/pull/132#issuecomment-576564457)

~**Note**: This PR updates all the charts (`enterprise-edition`, `push-proxy` and `team-edition`), let me know if this is something that you concern.~

#### Corresponding PRs

enterprise-edition: #134
push-proxy: #133

#### Ticket Link

Relates to #126 and mattermost/mattermost-server#13227

### DIFF
--- a/charts/mattermost-enterprise-edition/Chart.yaml
+++ b/charts/mattermost-enterprise-edition/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Mattermost Enterprise server with high availitibity.
 name: mattermost-enterprise-edition
-version: 1.1.0
+version: 1.0.1
 appVersion: 5.15.0
 keywords:
 - mattermost
--- a/charts/mattermost-enterprise-edition/Chart.yaml
+++ b/charts/mattermost-enterprise-edition/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Mattermost Enterprise server with high availitibity.
 name: mattermost-enterprise-edition
-version: 1.0.1
+version: 1.1.0
 appVersion: 5.15.0
 keywords:
 - mattermost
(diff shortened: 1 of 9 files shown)


khos2ow:
Bumped charts to next corresponding `minor` versions. I chose _minor_ over _patch_ because this is more like adding a new feature rather than fixing a thing. Please feel free to advise if the versioning is something you'd like to adjust.

khos2ow:
@jaydeland I bumped the chart versions to next available **minor** version.
Helm v3 support potentially is doable (there's a failure on CI because of that, but I haven't attempted to fix it as I wasn't sure if we want it at all or not)

jaydeland:
Yah - I am also not sure if we need - @cpanato??

> @jaydeland I bumped the chart versions to next available **minor** version.
> Helm v3 support potentially is doable (there's a failure on CI because of that, but I haven't attempted to fix it as I wasn't sure if we want it at all or not)

cpanato:
@khos2ow @jaydeland regarding helm v3
Since Helm 2 has a [1 year deprecation period](https://helm.sh/blog/2019-10-22-helm-2150-released/#helm-2-support-plan), I think will be easy for now to keep apiVersion v1 charts around till then (Nov 13, 2020).
Helm v3 can happily install APIVersion v1 charts, the only issue is with CRDs but we don't use that 😄 
So we follow the deprecation period and in Nov we move to V2

khos2ow:
@cpanato fair enough, I removed the changes for `enterprise-edition` and `push-proxy` from here and I will open corresponding PRs of their own.

Not supporting Helm v3 is fine, but I think it's worth double-checking that there is clear indication in the installation guide (or similar) that v3 is not supported.

khos2ow:
in that case hold off of merging this PR, let me quickly test with Helm v3 against a recent cluster (1.16, or 1.17). Because I'm using a builtin objects of the Helm (to make sure it didn't get deprecated going from v2 -> v3)

hanzei:
@jaydeland Gentle reminder to  review the changes when you are back from your weekend

PR #137, opened 2020-03-22
FIX: plugins tar and rm expression

hajowieland:
#### Summary
Fixes the extraction and removal of downloaded Plugin archives in the `extraInitContainers` section of the **Mattermost Enterprise** Helm chart.

Before, it tried to extract and remove the whole plugin URL - now it extracts the filename of the URL and uses only this part.


hajowieland:
had to sign the SLA - please delete this PR, sorry for any inconvenience caused

PR #144, opened 2020-05-21
Adding Ability to specify Service Account

jseiser:
#### Summary
Allows specifying a service account, name and annotations.  This will let you take advantage of things like https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html


charts/mattermost-enterprise-edition/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Mattermost Enterprise server with high availitibity.
 name: mattermost-enterprise-edition
-version: 1.1.1
+version: 1.1.2
```
cpanato, line 4:
```suggestion
version: 1.2.0
```
should be 1.2.0 since is adding a new feature

PR #186, opened 2020-09-16
Move to use context instead of env vars directly

cpanato:
#### Summary
Move to use context instead of env vars directly

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/MM-28110


cpanato:
will open using upstream

PR #192, opened 2020-10-29
fix hpa and node affinity

erezo9:


charts/mattermost-enterprise-edition/templates/deployment-mattermost-hpa.yaml
```diff
@@ -7,10 +7,10 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     helm.sh/chart:  {{ include "mattermost-enterprise-edition.chart" . }}
-  name: {{ include "mattermost-enterprise-edition.fullname" . }}}
+  name: {{ include "mattermost-enterprise-edition.fullname" . }}
 spec:
   scaleTargetRef:
-    apiVersion: {{ template "deployment.apiVersion" . }}
+    apiVersion: {{ template "mattermost-enterprise-edition.deployment.apiVersion" . }}
```
khos2ow, line 13: I understand that this is to fix a missing template name, but why was it added in the first place? I mean this can be hardocded `apps/v1` without the use of `template`?

khos2ow (commented):
Thanks @erezo9, looks good but I have one question below.

charts/mattermost-enterprise-edition/templates/deployment-mattermost-hpa.yaml
erezo9: @khos2ow 
I mean you can, buy if k8s will change the api - or lets say opensshift deployment, you need to compare semver

khos2ow: It's a `v1` stable release and the only change to this will be to go to `v2`. If that happens (if ever happens!), that's gonna be a breaking change for everything for all the helm charts in this repo (and else) and we need to track and mitigate it accordingly.

I'm fine with this use of `template` but it might give an impression that under some circumstance the `apiVersion` on this line won't be `apps/v1`, which is false.

Let's wait for @cpanato's feedback too.

erezo9: Okay, whatever it will be ill change it accordingly in the code

cpanato: most of the charts or use hardcoded or the previous way. I'm fine with that.
if v2 comes out we might need to release a new chart to fix other things

cpanato (approved):

PR #208, opened 2021-02-10
pod security context / persistent configuration changes

### LINKED ISSUE #100: [mattermost-team-edition] read only file system on start up
While installing the mattermost team edition helm chart on my GKE for the first time, i got a read only file system error without modifying anything : 

**Error message :** 

{"level":"error","ts":1563295457.9649801,"caller":"app/config.go:54","msg":"Failed to update config","error":"failed to persist: failed to write file: open /mattermost/config/config.json: read-only file system","errorVerbose":"open /mattermost/config/config.json: read-only file system

**My values.yaml file : **

# Image tag
image:
  tag: 5.12.1
# Persistense for mattermost
persistence:
  data:
    enabled: true
    size: 20Gi
    storageClass: standard
    accessMode: ReadWriteOnce

# Ingress params to access the cluster from outside
ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.class: nginx
    certmanager.k8s.io/cluster-issuer: XXX
  hosts:
    - XXX
  tls:
    - secretName: XXX-tls
      hosts:
        - XXX

# Mysql Params
mysql:
  enabled: true
  mysqlRootPassword: "Password"
  mysqlUser: "USer"
  mysqlPassword: "Password"
  mysqlDatabase: db_name

  repository: mysql
  tag: 5.7
  imagePullPolicy: IfNotPresent

  persistence:
    enabled: true
    storageClass: standard
    accessMode: ReadWriteOnce
    size: 10Gi

# Config File for mattermost
configJSON: {
  "ServiceSettings": {
    "SiteURL": "XXXX",
    "GoogleDeveloperKey": "XXXXX",
    "EnablePostUsernameOverride": true,
    "EnablePostIconOverride": true,
    "EnableLinkPreviews": true,
  },
  "TeamSettings": {
    "SiteName": "XXX",
    "EnableCustomBrand": true,
  },
  "EmailSettings": {
    "SendEmailNotifications": true,
    "RequireEmailVerification": true,
    "SMTPUsername": "XXX",
    "SMTPPassword": "XXX",
    "EnableSMTPAuth": true,
    "SMTPServer": "XXX",
    "SMTPPort": "587",
    "ConnectionSecurity": "STARTTLS",
    "EnableEmailBatching": true,
  },
  "PluginSettings": {
    "Enable": true,
    "EnableUploads": true,
  }
}

### LINKED ISSUE #203: [mattermost-team-edition] read only file system ONLY FOR >=5.30
When i run mattermost images >= 5.30 mattermost wont start due to `Error: failed to load configuration: failed to create store: unable to load on store creation: failed to persist: failed to write file: open /mattermost/config/config.json: read-only file system`

```
image:
  repository: mattermost/mattermost-team-edition
  tag: 5.31.0-rc2
  imagePullPolicy: IfNotPresent
```

When i switch to older versions like **5.29.1** mattermost will start.

mrparkers:
#### Summary

This PR allows users to configure the application pod's security context, which is required in order to persist changes to the filesystem for plugins and user data.

#### Ticket Link

Fixes #203

--------------------

Update - I ended up closing #209 and committed those changes here, as the pod security context changes are also required for the init container's ability to write the config json to a persistent volume with the appropriate user/group assignment.

Copy/paste of #209:

#### Summary

Before this change, there was no way to allow mattermost instances deployed via this chart to persist updates to its configuration.  This is because the configuration is controlled by helm, and is provided to the app using a read-only filesystem, which is a consequence of sourcing the config from a Kubernetes secret.  This means that tasks like installing and configuring a plugin is impossible or unnecessarily difficult, as the application has no way to persist config changes (as it cannot write to the read-only filesystem).

This PR introduces a new configuration setting, `persistence.config`, which controls whether or not mattermost will be allowed to update its own configuration.  If `persistence.config.enabled` is set to `true`, then the `configJSON` value will only be respected on initial helm install, and an init container will copy this JSON over to a writable persistent volume, which will then be used by the application.  This will allow the application to write any changes it needs to within the `config.json` file.

The downside of this is that further updates to `configJSON` after the initial install will do nothing.  Users that would prefer to manage their entire JSON config via helm can still do so by setting `persistence.config.enabled` to `false`.

A couple of notes:

- I didn't add documentation for this property, since it appears that `persistence.*` isn't documented.  I'm happy to add documentation for my changes as well as backfill docs for the other persistence variables as well.  I could also write an explicit section in the README about this change if you'd like.
- I set the default value of `persistence.config.enabled` to `true`, as this is the behavior I originally expected when I first used the chart.  I'm happy to update the default to `false` if you'd like.

#### Ticket Link

Fixes #100

### DIFF
--- a/charts/mattermost-enterprise-edition/Chart.yaml
+++ b/charts/mattermost-enterprise-edition/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Mattermost Enterprise server with high availitibity.
 name: mattermost-enterprise-edition
-version: 1.13.0
+version: 1.14.0
 appVersion: 5.29.0
 keywords:
 - mattermost
--- a/charts/mattermost-enterprise-edition/templates/deployment-mattermost-app.yaml
+++ b/charts/mattermost-enterprise-edition/templates/deployment-mattermost-app.yaml
@@ -49,6 +49,8 @@ spec:
       tolerations:
         {{ toYaml .Values.mattermostApp.tolerations | indent 8 }}
       {{- end }}
+      securityContext:
+        {{- toYaml .Values.mattermostApp.securityContext | nindent 8 }}
       {{ if .Values.serviceAccount.create }}
       serviceAccountName: {{ .Values.serviceAccount.name }}
       {{ end }}
--- a/charts/mattermost-enterprise-edition/values.yaml
+++ b/charts/mattermost-enterprise-edition/values.yaml
@@ -180,7 +180,9 @@ mattermostApp:
   ## Pod Security Context
   ## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
   securityContext:
-    fsGroup: 1000
+    fsGroup: 2000
+    runAsGroup: 2000
+    runAsUser: 2000
 
   resources: {}
     # limits:
--- a/charts/mattermost-enterprise-edition/values.yaml
+++ b/charts/mattermost-enterprise-edition/values.yaml
@@ -177,6 +177,11 @@ mattermostApp:
   # Ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
   tolerations: []
 
+  ## Pod Security Context
+  ## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+  securityContext:
+    fsGroup: 1000
+
   resources: {}
     # limits:
     #   cpu: 100m
--- a/charts/mattermost-team-edition/Chart.yaml
+++ b/charts/mattermost-team-edition/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Mattermost Team Edition server.
 name: mattermost-team-edition
-version: 3.22.0
+version: 3.23.0
 appVersion: 5.29.0
 keywords:
 - mattermost
--- a/charts/mattermost-team-edition/README.md
+++ b/charts/mattermost-team-edition/README.md
@@ -84,7 +84,9 @@ Parameter                             | Description                             
 `extraPodAnnotations`                 | Extra pod annotations to be used in the deployments                                             | `[]`
 `extraEnvVars`                        | Extra environments variables to be used in the deployments                                      | `[]`
 `extraInitContainers`                 | Additional init containers                                                                      | `[]`
-`securityContext.fsGroup`             | Sets the pod's filesystem group ID                                                              | `1000`
+`securityContext.fsGroup`             | Sets the pod's filesystem group ID                                                              | `2000`
+`securityContext.runAsUser`           | Sets the user ID for the pod                                                                    | `2000`
+`securityContext.runAsGroup`          | Sets the group ID for the pod                                                                   | `2000`
 `service.annotations`                 | Service annotations                                                                             | `{}`
 `service.loadBalancerIP`              | A user-specified IP address for service type LoadBalancer to use as External IP (if supported)  | `nil`
 `service.loadBalancerSourceRanges`    | list of IP CIDRs allowed access to load balancer (if supported)                                 | `[]`
--- a/charts/mattermost-team-edition/README.md
+++ b/charts/mattermost-team-edition/README.md
@@ -84,6 +84,7 @@ Parameter                             | Description                             
 `extraPodAnnotations`                 | Extra pod annotations to be used in the deployments                                             | `[]`
 `extraEnvVars`                        | Extra environments variables to be used in the deployments                                      | `[]`
 `extraInitContainers`                 | Additional init containers                                                                      | `[]`
+`securityContext.fsGroup`             | Sets the pod's filesystem group ID                                                              | `1000`
 `service.annotations`                 | Service annotations                                                                             | `{}`
 `service.loadBalancerIP`              | A user-specified IP address for service type LoadBalancer to use as External IP (if supported)  | `nil`
 `service.loadBalancerSourceRanges`    | list of IP CIDRs allowed access to load balancer (if supported)                                 | `[]`
--- a/charts/mattermost-team-edition/templates/deployment.yaml
+++ b/charts/mattermost-team-edition/templates/deployment.yaml
@@ -55,6 +55,27 @@ spec:
       {{- if .Values.extraInitContainers }}
       {{- .Values.extraInitContainers | toYaml | nindent 6 }}
       {{- end }}
+      {{- if .Values.persistence.config.enabled }}
+      - name: "init-config"
+        image: "{{ .Values.persistence.config.initContainerImage.repository }}:{{ .Values.persistence.config.initContainerImage.tag }}"
+        imagePullPolicy: {{ .Values.persistence.config.initContainerImage.imagePullPolicy }}
+        command:
+          - sh
+          - -c
+          - |
+            if test -f /persist-config/config.json; then
+              echo "config exists"
+            else
+              echo "creating initial config"
+              cp /init-config/config.json /persist-config/
+              chmod 0664 /persist-config/config.json
+            fi
+        volumeMounts:
+          - mountPath: "/init-config"
+            name: config-json
+          - mountPath: "/persist-config"
+            name: persistent-config
+      {{- end }}
       containers:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
@@ -82,9 +103,14 @@ spec:
             path: /api/v4/system/ping
             port: http
         volumeMounts:
+        {{- if .Values.persistence.config.enabled }}
+        - mountPath: /mattermost/config
+          name: persistent-config
+        {{- else }}
         - mountPath: /mattermost/config/config.json
           name: config-json
           subPath: config.json
+        {{- end }}
         - mountPath: /mattermost/data
           name: mattermost-data
         - mountPath: /mattermost/{{ trimPrefix "./" .Values.configJSON.PluginSettings.Directory }}
@@ -101,6 +127,11 @@ spec:
       - name: config-json
         secret:
           secretName: {{ include "mattermost-team-edition.fullname" . }}-config-json
+      {{- if .Values.persistence.config.enabled }}
+      - name: persistent-config
+        persistentVolumeClaim:
+          claimName: {{ .Values.persistence.config.existingClaim | default (include "mattermost-team-edition.fullname" .) }}-config
+      {{- end }}
       - name: mattermost-data
       {{ if .Values.persistence.data.enabled }}
         persistentVolumeClaim:
--- a/charts/mattermost-team-edition/templates/deployment.yaml
+++ b/charts/mattermost-team-edition/templates/deployment.yaml
@@ -43,6 +43,8 @@ spec:
       tolerations:
         {{ toYaml .Values.tolerations | indent 8 }}
       {{- end }}
+      securityContext:
+        {{- toYaml .Values.securityContext | nindent 8 }}
       initContainers:
       {{- if not .Values.externalDB.enabled }}
       - name: "init-mysql"
--- a/charts/mattermost-team-edition/templates/deployment.yaml
+++ b/charts/mattermost-team-edition/templates/deployment.yaml
@@ -75,6 +75,9 @@ spec:
             name: config-json
           - mountPath: "/persist-config"
             name: persistent-config
+        securityContext:
+          runAsUser: {{ .Values.securityContext.runAsUser }}
+          runAsGroup: {{ .Values.securityContext.runAsGroup }}
       {{- end }}
       containers:
       - name: {{ .Chart.Name }}
--- a/charts/mattermost-team-edition/templates/pvc-config.yaml
+++ b/charts/mattermost-team-edition/templates/pvc-config.yaml
@@ -1,0 +1,28 @@
+{{- if and .Values.persistence.config.enabled (not .Values.persistence.config.existingClaim) -}}
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: {{ include "mattermost-team-edition.fullname" . }}-config
+  labels:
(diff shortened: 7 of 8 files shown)


jfrerich:
@metanerd, is this something that you can take a look at?

khos2ow (commented):
From the functional changes in this PR, it LGTM, but I'm not entirely sure why it's needed. What has changed in 5.30 which caused this change to be needed?

mrparkers:
Hey @khos2ow, thanks for the reply.  To my knowledge, there weren't any changes within the mattermost application that warrant this suggested change I'm making.  I started working with mattermost a few weeks ago, and found these issues when trying to install the chart for the first time using the default settings.  I'll try and be a bit more clear about the problems I'm trying to solve.

## Problem 1:
When you run the chart as-is with the default settings, depending on the version of Kubernetes you're using (I'm using AWS EKS 1.16 here), you may end up with an install in which the `/mattermost/data` and `/mattermost/plugins` folders are owned by the `root` user/group, and are write protected.  This results in the following log message when the pod first starts:

```
{"level":"error","ts":1614276004.08384,"caller":"app/plugin.go:824","msg":"Failed to install prepackaged plugin","path":"/mattermost/prepackaged_plugins/mattermost-plugin-nps-v1.1.0-linux-amd64.tar.gz","error":"Failed to install extracted prepackaged plugin /mattermost/prepackaged_plugins/mattermost-plugin-nps-v1.1.0-linux-amd64.tar.gz: installExtractedPlugin: Unable to move plugin from temporary directory to final destination. Another plugin may be using the same directory name., mkdir plugins/com.mattermost.nps: permission denied","errorVerbose":"installExtractedPlugin: Unable to move plugin from temporary directory to final destination. Another plugin may be using the same directory name., mkdir plugins/com.mattermost.nps: permission denied\nFailed to install extracted prepackaged plugin /mattermost/prepackaged_plugins/mattermost-plugin-nps-v1.1.0-linux-amd64.tar.gz\ngithub.com/mattermost/mattermost-server/v5/app.(*App).processPrepackagedPlugin\n\tgithub.com/mattermost/mattermost-server/v5/app/plugin.go:876\ngithub.com/mattermost/mattermost-server/v5/app.(*App).processPrepackagedPlugins.func2\n\tgithub.com/mattermost/mattermost-server/v5/app/plugin.go:822\nruntime.goexit\n\truntime/asm_amd64.s:1373"}
```

If you get a shell to the running container (via `kubectl exec`), you can see that the filesystem looks like this:

```
~ $ pwd
/mattermost
~ $ ls -alh
total 232K
drwxr-sr-x    1 mattermo mattermo      40 Feb 25 18:03 .
drwxr-xr-x    1 root     root          46 Feb 25 17:59 ..
-rw-------    1 mattermo mattermo      15 Feb 25 18:03 .ash_history
-rw-r--r--    1 mattermo mattermo    1.2K Nov 13 18:25 MIT-COMPILED-LICENSE.md
-rw-r--r--    1 mattermo mattermo  197.3K Nov 13 18:25 NOTICE.txt
-rw-r--r--    1 mattermo mattermo    5.4K Nov 13 18:25 README.md
drwxr-xr-x    2 mattermo mattermo      53 Nov 13 18:44 bin
drwxr-xr-x    1 mattermo mattermo     196 Nov 13 18:44 client
drwxr-xr-x    2 mattermo mattermo      42 Feb 25 17:59 config
drwxr-xr-x    3 root     root        4.0K Feb 25 17:57 data
drwxr-xr-x    2 mattermo mattermo      44 Nov 13 18:44 fonts
drwxr-xr-x    2 mattermo mattermo     255 Nov 13 18:44 i18n
drwxr-xr-x    2 mattermo mattermo      28 Feb 25 17:59 logs
drwxr-xr-x    3 root     root        4.0K Feb 25 17:57 plugins
drwxr-xr-x    2 mattermo mattermo    4.0K Nov 13 18:45 prepackaged_plugins
drwxr-xr-x    2 mattermo mattermo    4.0K Nov 13 18:44 templates
~ $ whoami
mattermost
```

As you can see, these folders aren't going to be writable to the currently logged in user.  So this means that the default plugin install process on initial startup won't work, and installing new plugins won't work either, as seen here:

image

Which also produces the following log message:

```
{"level":"error","ts":1614276544.0832715,"caller":"mlog/log.go:229","msg":"Unable to move plugin from temporary directory to final destination. Another plugin may be using the same directory name.","path":"/api/v4/plugins/marketplace","request_id":"7qgu8tuzt3r37gut9uod14cnfc","ip_addr":"10.16.115.64","user_id":"e5bt3aj557rp3ccjogqh5nxgyy","method":"POST","err_where":"installExtractedPlugin","http_code":500,"err_details":"mkdir plugins/com.github.moussetc.mattermost.plugin.giphy: permission denied"}
```

This also causes issues with persisting user profile data, as seen by the following log message after a user registers:

```
{"level":"error","ts":1614276521.279308,"caller":"mlog/log.go:229","msg":"Encountered an error creating the directory for the new file.","path":"/api/v4/users/e5bt3aj557rp3ccjogqh5nxgyy/image","request_id":"nk5wokdjsfgjzpc1e8ms6diama","ip_addr":"10.16.115.64","user_id":"e5bt3aj557rp3ccjogqh5nxgyy","method":"GET","err_where":"WriteFile","http_code":500,"err_details":"directory=/mattermost/data/users/e5bt3aj557rp3ccjogqh5nxgyy, err=mkdir data/users: permission denied"}
```

---

After the `securityContext` changes proposed in this PR, the filesystem instead looks like this (notice the group change for the `plugins` and `data` folders):

```
~ $ pwd
/mattermost
~ $ ls -alh
total 232K
drwxr-sr-x    1 mattermo mattermo      40 Feb 25 20:10 .
drwxr-xr-x    1 root     root          46 Feb 25 20:08 ..
-rw-------    1 mattermo mattermo      12 Feb 25 20:10 .ash_history
-rw-r--r--    1 mattermo mattermo    1.2K Nov 13 18:25 MIT-COMPILED-LICENSE.md
-rw-r--r--    1 mattermo mattermo  197.3K Nov 13 18:25 NOTICE.txt
-rw-r--r--    1 mattermo mattermo    5.4K Nov 13 18:25 README.md
drwxr-xr-x    2 mattermo mattermo      53 Nov 13 18:44 bin
drwxr-xr-x    1 mattermo mattermo     196 Nov 13 18:44 client
drwxr-xr-x    2 mattermo mattermo      42 Feb 25 20:08 config
drwxrwsr-x    4 root     mattermo    4.0K Feb 25 20:08 data
drwxr-xr-x    2 mattermo mattermo      44 Nov 13 18:44 fonts
drwxr-xr-x    2 mattermo mattermo     255 Nov 13 18:44 i18n
drwxr-xr-x    2 mattermo mattermo      28 Feb 25 20:08 logs
drwxrwsr-x    4 root     mattermo    4.0K Feb 25 20:08 plugins
drwxr-xr-x    2 mattermo mattermo    4.0K Nov 13 18:45 prepackaged_plugins
drwxr-xr-x    2 mattermo mattermo    4.0K Nov 13 18:44 templates
~ $ whoami
mattermost
```

Which allows the plugin to be installed successfully:

image


And, the user data is persisting as expected.

## Problem 2:
With the fixes described above, the `gif` plugin can be installed, but it can not be configured from the UI:

image

This emits the following log message:

```
{"level":"error","ts":1614284051.381758,"caller":"mlog/log.go:229","msg":"An error occurred saving the configuration.","path":"/api/v4/config","request_id":"1jzb9x83ztb5i88j4dbjs4g7ew","ip_addr":"10.16.119.117","user_id":"y84dbz4ziibwjndtf3ofpmdtma","method":"PUT","err_where":"saveConfig","http_code":500,"err_details":"failed to persist: failed to write file: open /mattermost/config/config.json: read-only file system"}
```

This is happening because the volume containing the `config.json` file is being created from a Kubernetes secret, resulting in a read-only filesystem.  While the docs suggest that you can mount this volume with any mode you want (perhaps `0644`), this issue indicates that this does not appear to be true: https://github.com/kubernetes/kubernetes/issues/62099.  When mounting from a secret, the `config.json` file in the filesystem will always look like this:

```
~/config $ pwd
/mattermost/config
~/config $ ls -alh
total 12K
drwxr-xr-x    2 mattermo mattermo      42 Feb 25 20:47 .
drwxr-sr-x    1 mattermo mattermo      40 Feb 25 20:48 ..
-rw-r--r--    1 mattermo mattermo     243 Nov 13 18:25 README.md
-rw-r--r--    1 root     mattermo    5.5K Feb 25 20:45 config.json
```

So, the only way to allow the mattermost application to modify this file is to change the way we're persisting it - using a volume that can be mounted as a writable filesystem.  The approach in this PR checks if `persistence.config.enabled` is set to `true` - if it is, we'll treat the `configJSON` value as an initial config, and use an `initContainer` to write this config to a new volume, and attach that volume to the mattermost application pod instead of attaching the volume from the secret.  This results in a `/mattermost/config/config.json` file that can be modified by the application, as seen here:

```
~/config $ pwd
/mattermost/config
~/config $ ls -alh
total 40K
drwxrwsr-x    3 root     mattermo    4.0K Feb 25 20:56 .
drwxr-sr-x    1 mattermo mattermo      40 Feb 25 20:58 ..
-rw-rw-r--    1 mattermo mattermo   18.0K Feb 25 20:56 config.json
drwxrws---    2 root     mattermo   16.0K Feb 25 20:54 lost+found
```

Which then allows the plugin to be installed and configured from the UI.

However, this approach comes at a cost - as the `configJSON` value will only be respected on the initial helm install.  Users won't be able to use helm to further configure the application, only the UI.  I personally think this trade-off is alright, as long as it's properly documented (which I'm happy to do myself in order to get this PR merged).

Alternatively, a more sophisticated approach might include some sort of sidecar container that keeps track of changes on either side (helm or UI), and merges those changes appropriately.  This might be similar to the approach taken by the Jenkins chart when using the "configuration as code" plugin: https://github.com/jenkinsci/helm-charts/tree/main/charts/jenkins#config-as-code-with-or-without-auto-reload.  Although I don't really have the bandwidth to work on something like this.

---

I hope that helps clear things up, please let me know if you have any other questions.

agusl88:
I just tested this PR on my installation and worked perfectly, the config.json was initializated from the secrets and copied into the new persistentVolume and now all settings are writable from the System Console. 

The plugin marketplace also works and persist plugins withouth any issues.

khos2ow (commented):
Thank you @mrparkers for the in-depth explanation and apologize for late response. I'm completely fine with the your proposed changed, however

> However, this approach comes at a cost - as the configJSON value will only be respected on the initial helm install. Users won't be able to use helm to further configure the application, only the UI

this seems to be a bit alarming, and definitely needs more eyes on it.

/cc @stylianosrigas @cpanato @jwilander

mrparkers:
Appreciate the response @khos2ow.  I just want to reiterate that the config persistence changes are only optional - the default value for config persistence is the functionality that's currently in master.  If it would help, I can document this change in the README, and clarify the tradeoff that comes with opting into this functionality.

For what it's worth, I've been using this version of the chart for a few weeks now, and I personally think it's a better experience managing these settings in the UI or even the Mattermost API rather than manually editing the `config.json` file (especially for plugin settings).

Anyways, please let me know if there's any way I can help move this along.  Thanks! 🚀

khos2ow:
That's great and I completely understand where you're coming from on this matter, but I, personally, am not in a position to call a final shot of this matter, specially that this scopes a bit beyond the chart and deployment itself and moves a bit into underlying config management of Mattermost as a whole.

agusl88:
Maybe setting the toggle off by default is the best approach, since it wont forcefully change the way of handling the config.json for people that is already using the chart.

Imo the ideal solution would be to move the plugins configuration out of the config.json, so we can manage the server settings via the helm-deployment while having the plugin marketplace functional, but i believe that's a suggestion that would have to be done to the core repository. In the meantime, this change seems like a good workaround to avoid breaking the plugin marketplace for everyone using this chart withouth giving them an easy way to fix the problem

Blackclaws:
Why is this still not merged if I may ask? I just today had the same issues and had to manually change the group of the mounted volumes as there is no default security context set. 

The read-only config is also quite problematic and even trying to migrate it into the database configuration as described in the documentation is not possible as the migration fails because the config is readonly

sotolucas:
hey there @mrparkers !
In some cases, the config persistence size value of 50MiB seem not to work.
When I set it to 1Gi for example, it works fine.
I'm using DigitalOcean Kubernetes Service.
Anyway that's the only thing I would change, but apart from that everything else LGTM.

cpanato (changes requested):
thanks for this change

can you split the PR in two? one for TE and another for EE

thanks!

ahmaddanialmohd:
A customer mentioned that they were able to solve it by creating a separate mount point. A question that they have here is whether it is possible to temporarily apply **extraInitContainers** like in the following link:

* [[mattermost-team-edition] read only file system on start up #100
](https://github.com/mattermost/mattermost-helm/issues/100)

syuo7:
Hello.
When plan this PR be merged?
I'm looking forward to resolving this issue.

Thank you.

jasonblais:
@cpanato why do we want to split the PR into two, any harm in keeping it in one and reviewing/merging?

cpanato:
@jasonblais because the automation for the release of the charts only supports one chart being updated at time per PR, and also helps to track down the changes

jasonblais:
@cpanato I split out the changes to
 - enterprise-edition: https://github.com/mattermost/mattermost-helm/pull/223
 - team-edition: https://github.com/mattermost/mattermost-helm/pull/224

@mrparkers Huge thanks for your contribution! I hope it was okay for me to split your changes into the above two PRs. Currently, our automation for the release of the charts only supports one chart being updated at time per PR. If you'd prefer to make the split yourself, feel free to do so, and I'll close mine.

Hope we get these excellent changes merged soon. :)

cpanato:
closing this PR in favor or the ones that Jason opened

mrparkers:
No problem at all, I didn't have the bandwidth to redo this so I'm happy these changes were able to make their way in somehow.  Thanks for the initiative!

PR #209, opened 2021-02-11
[mattermost-team-edition] add configuration to persist mattermost config json

### LINKED ISSUE #100: [mattermost-team-edition] read only file system on start up

mrparkers:
#### Summary

Before this change, there was no way to allow mattermost instances deployed via this chart to persist updates to its configuration.  This is because the configuration is controlled by helm, and is provided to the app using a read-only filesystem, which is a consequence of sourcing the config from a Kubernetes secret.  This means that tasks like installing and configuring a plugin is impossible or unnecessarily difficult, as the application has no way to persist config changes (as it cannot write to the read-only filesystem).

This PR introduces a new configuration setting, `persistence.config`, which controls whether or not mattermost will be allowed to update its own configuration.  If `persistence.config.enabled` is set to `true`, then the `configJSON` value will only be respected on initial helm install, and an init container will copy this JSON over to a writable persistent volume, which will then be used by the application.  This will allow the application to write any changes it needs to within the `config.json` file.

The downside of this is that further updates to `configJSON` after the initial install will do nothing.  Users that would prefer to manage their entire JSON config via helm can still do so by setting `persistence.config.enabled` to `false`.

A couple of notes:

- I didn't add documentation for this property, since it appears that `persistence.*` isn't documented.  I'm happy to add documentation for my changes as well as backfill docs for the other persistence variables as well.  I could also write an explicit section in the README about this change if you'd like.
- I set the default value of `persistence.config.enabled` to `true`, as this is the behavior I originally expected when I first used the chart.  I'm happy to update the default to `false` if you'd like.

#### Ticket Link

Fixes #100

### DIFF
--- a/charts/mattermost-team-edition/Chart.yaml
+++ b/charts/mattermost-team-edition/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Mattermost Team Edition server.
 name: mattermost-team-edition
-version: 3.22.0
+version: 3.23.0
 appVersion: 5.29.0
 keywords:
 - mattermost
--- a/charts/mattermost-team-edition/templates/deployment.yaml
+++ b/charts/mattermost-team-edition/templates/deployment.yaml
@@ -53,6 +53,27 @@ spec:
       {{- if .Values.extraInitContainers }}
       {{- .Values.extraInitContainers | toYaml | nindent 6 }}
       {{- end }}
+      {{- if .Values.persistence.config.enabled }}
+      - name: "init-config"
+        image: "{{ .Values.persistence.config.initContainerImage.repository }}:{{ .Values.persistence.config.initContainerImage.tag }}"
+        imagePullPolicy: {{ .Values.persistence.config.initContainerImage.imagePullPolicy }}
+        command:
+          - sh
+          - -c
+          - |
+            if test -f /persist-config/config.json; then
+              echo "config exists"
+            else
+              echo "creating initial config"
+              cp /init-config/config.json /persist-config/
+              chmod 0664 /persist-config/config.json
+            fi
+        volumeMounts:
+          - mountPath: "/init-config"
+            name: config-json
+          - mountPath: "/persist-config"
+            name: persistent-config
+      {{- end }}
       containers:
       - name: {{ .Chart.Name }}
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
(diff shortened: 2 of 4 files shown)


mrparkers:
Closing in favor of #208

PR #210, opened 2021-02-23
Upgrade MM-TE to be helm v3 fully compatible

cpanato:
#### Summary
Helm 2 is deprecated, and we should not support more this version, at this stage whoever is using it should have already migrated from v2 to v3.
This PR update the chart to the helm v3 way 😄 

- Remove the owners names, because this was set up when this chart was in the helm/charts repo, not needed here for now
- update the dependencies file to the helm v3 pattern
- update the ct config to ignore the maintainers in the chart.yaml file
- bump MM-TE to the latest version

tested with a previous chart version and upgraded to these changes and no issues found during the upgrade.

after this will do the  same for the other charts too

cc @stylianosrigas @khos2ow @angeloskyratzakos 

#### Ticket Link

NONE

### DIFF
--- a/charts/mattermost-team-edition/Chart.yaml
+++ b/charts/mattermost-team-edition/Chart.yaml
@@ -1,8 +1,9 @@
-apiVersion: v1
+apiVersion: v2
 description: Mattermost Team Edition server.
+type: application
 name: mattermost-team-edition
-version: 3.22.0
-appVersion: 5.29.0
+version: 4.0.0
+appVersion: 5.32.1
 keywords:
 - mattermost
 - communication
--- a/charts/mattermost-team-edition/Chart.yaml
+++ b/charts/mattermost-team-edition/Chart.yaml
@@ -12,3 +12,8 @@ icon: http://www.mattermost.org/wp-content/uploads/2016/04/icon.png
 sources:
 - https://github.com/mattermost/mattermost-server
 - https://github.com/mattermost/mattermost-helm
+dependencies:
+- name: mysql
+  version: 1.6.4
+  repository: https://charts.helm.sh/stable
+  condition: mysql.enabled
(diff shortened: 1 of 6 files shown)


angeloskyratzakos (approved):
The legend is still around! LGTM! Thank you @cpanato!

stylianosrigas (approved):
Great stuff, thanks @cpanato 🚀

khos2ow (approved):
LGTM! But one question why only TE though? don't we need it for the other two (or at least EE)?

cpanato:
@khos2ow yes, but the TE is the open-source one :) so starting with that 
I can follow up in the EE later, but for an open-source perspective I'm interested in the TE one first 😄

khos2ow:
Yeah makes sense! Feel free to proceed with this :+1: :rocket:

cpanato:
with the power that was given to me, I will merge this PR 😂

cpanato:
i will follow up the other charts as well later this week/next week

charts/mattermost-team-edition/Chart.yaml
```diff
@@ -1,8 +1,9 @@
-apiVersion: v1
+apiVersion: v2
 description: Mattermost Team Edition server.
+type: application
 name: mattermost-team-edition
-version: 3.22.0
-appVersion: 5.29.0
+version: 4.0.0
+appVersion: 5.32.1
```
FlipEnergy, line 6: Sorry I'm a little late but I just installed 4.0.0 and had no problems, but I noticed there's a tag value set to 5.29.0, not this appVersion. If the tag is updated to match this newer app version, this issue comes up https://github.com/mattermost/mattermost-helm/issues/203

cpanato:
@FlipEnergy thanks and a good catch, I missed updating the version of MM in the values.yaml 🤦 

will do when having a free cycle

PR #211, opened 2021-02-25
mm-te: update values to use latest MM-TE available

cpanato:
#### Summary
my fault and forgot to update the values.yaml from this PR https://github.com/mattermost/mattermost-helm/pull/210

#### Ticket Link
none


cpanato:
need more changes looks like after 5.30 it write things back to the config....

PR #213, opened 2021-03-20
Fix Mattermost 5.30+ startup issue caused by read only config.json

### LINKED ISSUE #203: [mattermost-team-edition] read only file system ONLY FOR >=5.30

xtermi2:
#### Summary
This PR will copy the config.json from the k8s secret, which is of course read only, to a k8s emptyDir via a initContainer.
Then the matermost container can use the config.json from the emptyDir. The config must be writeable since Mattermost 5.30+ because Mattermost will pretty-print the config.json since this version.

In my opinion, this PR is a better option than #208 becaus you will always get a reproduceable container at startup and you can enforce/change stuff via infrastructure-as-code. And it should be 100% compatible with older versions and you don't need a new PVC

#### Ticket Link
This PR will fix #203

### DIFF
--- a/charts/mattermost-team-edition/Chart.yaml
+++ b/charts/mattermost-team-edition/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: Mattermost Team Edition server.
 type: application
 name: mattermost-team-edition
-version: 4.0.0
+version: 4.0.1
 appVersion: 5.32.1
 keywords:
 - mattermost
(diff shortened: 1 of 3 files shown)


cpanato:
@xtermi2 thanks for your PR, Since some releases ago Mattermost server supports the configuration to be managed in the DB. I think that will be the right approach to update the TE chart.

I will work in the refactor of the TE chart to fix this issue

For now I will close this PR, thanks so much for this!

PR #221, opened 2021-05-14
Focalboard Helm Chart

stafot:
If this commit applied, people will be able to install focalboard as a helm chart
Issue: https://github.com/mattermost/focalboard/issues/234

#### Summary
Adds a helm chart for focalboard

#### Ticket Link
mattermost/focalboard#234

### DIFF
--- a/charts/focalboard/templates/deployment.yaml
+++ b/charts/focalboard/templates/deployment.yaml
@@ -35,9 +35,6 @@ spec:
             - name: http
               containerPort: 8000
               protocol: TCP
-            - name: metrics
-              containerPort: 9092
-              protocol: TCP
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
--- a/charts/focalboard/templates/deployment.yaml
+++ b/charts/focalboard/templates/deployment.yaml
@@ -35,6 +35,9 @@ spec:
             - name: http
               containerPort: 8000
               protocol: TCP
+            - name: metrics
+              containerPort: 9092
+              protocol: TCP
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
--- a/charts/focalboard/templates/deployment.yaml
+++ b/charts/focalboard/templates/deployment.yaml
@@ -47,6 +47,17 @@ spec:
               port: http
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          volumeMounts:
+          - name: data
+            mountPath: /var/lib/mysql
+      volumes:
+        - name: data
+          {{- if .Values.persistence.enabled }}
+          persistentVolumeClaim:
+            claimName: {{ template "focalboard.fullname" . }}
+          {{- else }}
+          emptyDir: {}
+      {{- end -}}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
--- a/charts/focalboard/templates/deployment.yaml
+++ b/charts/focalboard/templates/deployment.yaml
@@ -1,0 +1,61 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "focalboard.fullname" . }}
+  labels:
+    {{- include "focalboard.labels" . | nindent 4 }}
+spec:
+  {{- if not .Values.autoscaling.enabled }}
+  replicas: {{ .Values.replicaCount }}
+  {{- end }}
+  selector:
+    matchLabels:
+      {{- include "focalboard.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "focalboard.selectorLabels" . | nindent 8 }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "focalboard.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - name: http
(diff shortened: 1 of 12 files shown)


jjlakis:
I can confirm that this just worked :) . I think that being able to create a PVC to persist data inside the chart would be important thing to add.

stafot:
@jjlakis Thanks for your comment, I will add this too.

charts/focalboard/templates/deployment.yaml
```diff
@@ -0,0 +1,72 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "focalboard.fullname" . }}
+  labels:
+    {{- include "focalboard.labels" . | nindent 4 }}
+spec:
+  {{- if not .Values.autoscaling.enabled }}
```
spirosoik, line 8: I think autoscaling is not the proper wording as without scaling policies etc right? I think just pass the `replicaCount` without if.

charts/focalboard/templates/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - name: http
+              containerPort: 8000
+              protocol: TCP
+          livenessProbe:
+            httpGet:
+              path: /
```
spirosoik, line 42: I think this should be configurable and also we need to have a proper `/health` endpoint, something which the focalboard team can add or we can add it. Also there is a great post about liveness and rediness so we cannot kill the app. https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-yourself-in-the-foot

charts/focalboard/templates/deployment.yaml
```diff
@@ -0,0 +1,72 @@
...
+              port: http
+          readinessProbe:
+            httpGet:
+              path: /
+              port: http
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+          volumeMounts:
+          - name: data
+            mountPath: /var/lib/mysql
```
spirosoik, line 52: could you elaborate? I mean is there a database in the container?

charts/focalboard/templates/pvc.yaml
```diff
@@ -0,0 +1,24 @@
+
```
spirosoik, line 1: let's remove the line here

charts/focalboard/values.yaml
```diff
@@ -0,0 +1,93 @@
...
+  enabled: false
+  annotations: {}
+    # kubernetes.io/ingress.class: nginx
+    # kubernetes.io/tls-acme: "true"
+  hosts:
+    - host: focalboard.example.com
+      paths:
+      - path: /
+        backend:
+          serviceName: chart-example.local
```
spirosoik, line 53: let's rename it to focalboard oriented name as default value

spirosoik (commented):
a few minor changes

charts/focalboard/templates/deployment.yaml
stafot: typo: I ment to have /data. Nice catch thanks

stafot:
@Hokwang The default dbtype is sqllite, thus as follow-up improvement after @jjlakis suggestion, added persistence on volume to avoid losing data after pod's restart. 
So to answer your question, no postgres is required although postgres is the recommended db to use [ref](https://www.focalboard.com/download/personal-edition/ubuntu/#install-postgresql-recommended). This Helm chart is the first iteration and an MVP. I would love in the near future to add more functionalities and possibilities.

charts/focalboard/templates/deployment.yaml
stafot: Yeap remove it for now and let's discuss it with `focalboard` team to have proper endpoints. Thanks for pointing this out.

stafot:
@Hokwang for sure next step would be to support postgresql either as k8s service or as external resource (f.e RDS endpoint). And since Focalboard, soon will add support for MySQL we most probably add support for MySQL too. But both of them are beyond the scope of this first PR. We will iterate on them soon though.

chenilim:
Thanks @stafot! Sorry for the delay. I don't have much (if any) experience with Helm charts, so can't fully review it, but the description and basic parameters look right to me.

Question: We want to expose environment variables to Docker (i.e. all the settings in [config.go](https://github.com/mattermost/focalboard/blob/main/server/services/config/config.go)). This includes the dbtype. Do those need to be specified in the dockerfile and / or the helm chart?

@jespino is semi-out this week. FYI @sbishel, is there anyone else who should review this? Thanks.

stafot:
@chenilim Yes, in my understanding focalboard need to support customizations through environment variables which can be used by the Helm chart.

spirosoik (approved):

jespino:
@chenilim yes we want to suppport environment variables for sure, that is something that is supported in mattermost and is going to be needed sooner than later. Probably one of the things that we want is to extract common parts from the current existing config architecture from mattermost and replace our quick-and-dirty solution for that. Mattermost configuration support environment variables and database storage, for example, to features really needed for the cloud environment.

@stafot I have one concern about the helm chart by itself, the helm chart is to run this in kubernetes, and running something in kubernnetes that doesn't support multi-node (HA) is kinda meh, so. Probably we are not going to support focalboard multi-node feature except as part of the mattermost-focalboard combination, so... I'm not sure if this makes sense now, or we want to think about HA clustering in standalone focalboard.

stafot:
@jespino It would be nice to be HA, but it is not required. Kubernetes even with single point of failure if you support external data store or persistent volume as we did in this PR, supports some resilience (you don't lose your data on a downtime). 
For me make sense to have a helm a) because people already asked for it to try it on their clusters, b) Is an easy out of the box installation for many people. Anyway is totally Focalboard team's call if you want this helm chart to be offered or not. So feel free to close this PR if you think that is not worth to offer a helm chart.
For the environment variables when Focalboard supports some of them we could follow-up some PRs to add them on the helm chart too.

jespino:
@stafot if you see value on the helm great, can we just restrict the replicaCount to one for the focalboard app?

spirosoik (commented):
@stafot one more comment we expose prometheus in focalboard so we can add the prometheus port and service too check this commit https://github.com/mattermost/focalboard/commit/8f746a44bd0a2b390e82e865df78b88ee3d98c96#diff-b4fc4f7ae1b99a6ebbed10408d87022ce921d95b1c015946277ae200cb262ff2

stafot:
@spirosoik Seems that right now is not exposed the port `9092` on docker container of focalboard. I will merge it without this change and I will raise a PR to fix this issue on focalboard repo.

PR #227, opened 2021-06-17
Fix a typo in documentation

oh6hay:
I assume that is a typo


cpanato (changes requested):
thanks for fixing this

can you bump the chart version?

oh6hay:
sure, like this?

cpanato (approved):

PR #252, opened 2021-09-28
Add a helm chart for Chaos Engine Mattermost application

spirosoik:
#### Summary
Chaos Engine is a Mattermost Application which is used to create gamedayys tight coupled with
Mattermost as collaboration tool. Helm chart will help deploy in a kubernetes cluster.

#### Ticket Link
Ticket: https://mattermost.atlassian.net/browse/MM-38883

### DIFF
--- a/charts/mattermost-chaos-engine/values.yaml
+++ b/charts/mattermost-chaos-engine/values.yaml
@@ -1,0 +1,90 @@
+# Default values for mattermost-chaos-engine.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: mattermost/mattermost-app-chaosengine
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "latest"
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+
+podAnnotations: {}
+
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext: {}
+  # capabilities:
+  #   drop:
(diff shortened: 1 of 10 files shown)


charts/mattermost-chaos-engine/values.yaml
```diff
@@ -0,0 +1,90 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: mattermost/mattermost-app-chaosengine
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "latest"
```
stafot, line 11: @spirosoik Ideally we don't want latest tag.

charts/mattermost-chaos-engine/values.yaml
```diff
@@ -0,0 +1,90 @@
...
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+  # limits:
+  #   cpu: 100m
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+autoscaling:
```
stafot, line 79: This block is not needed probably

spirosoik: why not let's keep it

charts/mattermost-chaos-engine/values.yaml
spirosoik: as default value it's fine right? typically it shouls respect chart app version if you don't pass the tag at all

spirosoik:
@stafot @stylianosrigas @angeloskyratzakos @Mshahidtaj  please have a look

stafot:
@spirosoik Did you understand why circleCI fails?

stafot (approved):

spirosoik:
> @spirosoik Did you understand why circleCI fails?

yes was about test connection and 404 of wget when chaos engine run. So I used a URL which always exist to get 200 and it's representative. Check my last commit

Mshahidtaj (approved):
LGTM

PR #266, opened 2021-10-26
Add option for ingressClassName for 1.18 and grate k8s versions

erezo9:
#### Summary
Adds option to retrieve ingressClassName since 1.18 removes annotation option


emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @erezo9

PR #294, opened 2022-01-17
Bump CRD's to new version

patatman:
#### Summary

This bumps the CRD's to the latest version of the operator.
I had some missing settings and config during install of the Helm chart.


patatman:
Looks like it needs some more work then I expected. 
I'll have a look tomorrow. My bad

PR #295, opened 2022-01-18
Bump CRD's

patatman:
#### Summary

This bumps the CRD's to the latest version of the operator.
I had some missing settings and config during install of the Helm chart.

### DIFF
--- a/charts/mattermost-operator/Chart.yaml
+++ b/charts/mattermost-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: mattermost-operator
 description: A Helm chart for Mattermost Operator
 type: application
-version: 0.2.1
+version: 0.2.2
 appVersion: 1.16.0
 keywords:
 - operator
--- a/charts/mattermost-operator/crds/crd-clusterinstallations.yaml
+++ b/charts/mattermost-operator/crds/crd-clusterinstallations.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.5.0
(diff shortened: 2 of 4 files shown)


Szymongib (approved):
LGTM! Thanks for updating it @patatman

patatman:
One thing I did notice, I see a Kustomize file in the original operator repo. Is this used in the files in any way? I could only find a reference in the makefile, but not inside the CRD's itself.

Just double-checking.

Szymongib:
> One thing I did notice, I see a Kustomize file in the original operator repo. Is this used in the files in any way? I could only find a reference in the makefile, but not inside the CRD's itself.

It is not used for CRDs only to create full installation yaml.

patatman:
I just noticed i'm getting an error when running Helm upgrade with this newer version: 
```
Error: UPGRADE FAILED: cannot patch "mattermost-operator" with kind Deployment: Deployment.apps "mattermost-operator" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"mattermost-operator", "app.kubernetes.io/managed-by":"Helm", "app.kubernetes.io/name":"mattermost-operator", "helm.sh/chart":"mattermost-operator-0.2.2", "name":"mattermost-operator"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

Szymongib:
It looks like we are using `helm.sh/chart` in a selector, which is immutable. The value is using chart version so changing it produces an error https://github.com/mattermost/mattermost-helm/blob/master/charts/mattermost-operator/templates/mattermost-operator/deployment.yaml#L17.

We will have to clean up the selector in the deployment as this will cause issues every time we bump the chart version. I think we can do it separately tho.

For now, you will have to recreate the deployment when doing the Helm upgrade.

stylianosrigas (approved):
LGTM, thanks @patatman

PR #297, opened 2022-01-26
Do not pass Chart version to deployment selectors

Szymongib:
#### Summary

Currently, the Chart version is passed to the value of one of the selector labels.
The selector, however, is immutable therefore any bump to the chart version makes it impossible to upgrade the chart with `helm upgrade`.

This PR removes the` helm.sh/chart` label from selector to fix this issue.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-41087

### DIFF
--- a/charts/mattermost-operator/Chart.yaml
+++ b/charts/mattermost-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: mattermost-operator
 description: A Helm chart for Mattermost Operator
 type: application
-version: 0.2.2
+version: 0.3.0
 appVersion: 1.16.0
 keywords:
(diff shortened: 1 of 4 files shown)


cpanato (commented):
did you test the upgrade? To check if nothing will be lost?

Szymongib:
> did you test the upgrade? To check if nothing will be lost?

As mentioned in the description upgrades do not work right now because of the selector, so for this change, you cannot upgrade either (deployment need to be manually deleted).

This PR should fix upgrades going forward.

cpanato (approved):

stylianosrigas (approved):
LGTM

PR #323, opened 2022-04-08
Fix service account template

jonathanwiemers:
This fixes an issue with creating the service account. The current code creates `namespace:mynamespace` which leads to yaml errors

### DIFF
--- a/charts/mattermost-enterprise-edition/templates/service-account.yaml
+++ b/charts/mattermost-enterprise-edition/templates/service-account.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ .Values.serviceAccount.name }}
-  namespace: {{- .Release.Namespace -}}
+  namespace: {{ .Release.Namespace -}}
 {{- if .Values.serviceAccount.annotations }}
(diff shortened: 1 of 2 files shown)


charts/mattermost-enterprise-edition/templates/service-account.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ .Values.serviceAccount.name }}
-  namespace: {{- .Release.Namespace -}}
+  namespace: {{ .Release.Namespace -}}
```
Szymongib, line 6: Just to avoid confusion in the future, I would make it consistent with other values.
```suggestion
  namespace: {{ .Release.Namespace }}
```

Szymongib (commented):
Thanks for the fix.

Please bump the chart version in `Chart.yaml` to 2.6.3 to make CI happy, otherwise LGTM.

spirosoik:
+1 what @Szymongib mentioned

jonathanwiemers:
@Szymongib fixed

Szymongib (approved):
Thanks! 
LGTM

PR #329, opened 2022-04-19
fix-extraEnv-default-value

ruckc:
#### Summary
I messed up the default value for extraEnv in #316. This fixes the jobserver.extraEnv to be an array, which is what the template is expecting

Without this change, if trying to use jobserver's extraEnvs, you get this error:

```
cannot overwrite table with non table for mattermost-enterprise-edition.global.features.jobserver.extraEnv
```

### DIFF
--- a/charts/mattermost-enterprise-edition/values.yaml
+++ b/charts/mattermost-enterprise-edition/values.yaml
@@ -65,7 +65,7 @@ global:
       # Ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
       tolerations: []
       # Allows the specification of additional environment variables for Mattermost Jobserver
-      extraEnv: {}
+      extraEnv: []
     notifications:
       # Push proxy must be configured or useHPNS must be true for push noticiations to work.
       push:


ruckc:
the tests need to be retried

ruckc:
@spirosoik / @pfltdv / @stylianosrigas - can you review?

pfltdv (approved):

spirosoik:
@pfltdv did you give a try to test locally?

pfltdv:
> @pfltdv did you give a try to test locally?

@spirosoik Yes tested with and without jobserver extra vars. Had no issue at both executions.

spirosoik (approved):

phoinixgrr (commented):
@ruckc  

Fine with merging this. 

Would you mind also commenting @ `values.yaml` some examples on how the Jobserver `extraEnv` directive can be used? 
For Example: 
https://github.com/mattermost/mattermost-helm/blob/master/charts/mattermost-enterprise-edition/values.yaml#L245-L251

jonathanwiemers:
@ruckc any reason this wasn't merged / updated anymore? Just saw that i've just create [a PR](https://github.com/mattermost/mattermost-helm/pull/348) for the same issue.

ruckc:
@jonathanwiemers - no real reason, other than the 3 week lag on feedback.  My need for this was removed as we moved away from mattermost due to the integration and license headaches.

PR #353, opened 2022-07-19
Add DNSConfig customization ability to RTCD service chart

stafot:
Issue: CLD-3730
Signed-off-by: Stavros Foteinopoulos <stafot@gmail.com>

#### Summary
Add DNSConfig customization ability to RTCD service chart

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-3730

### DIFF
--- a/charts/mattermost-rtcd/Chart.yaml
+++ b/charts/mattermost-rtcd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: mattermost-rtcd
 description: A Helm chart for Kubernetes to deploy Mattermost's RTCD
 type: application
-version: 1.0.0
+version: 1.1.0
 appVersion: latest
 keywords:
 - mattermost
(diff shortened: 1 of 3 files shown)


streamer45:
Not sure what's the impact of this but to share some context, from the plugin side we currently run a resolution on the rtcd host every 10 seconds to check if new hosts came up/went down in order to support horizontal scaling.

streamer45 (approved):
> @streamer45 Yeap, this lookup every 10 seconds with the default `ndots` configuration which defaults to 5 on kubernetes, creates 10 queries while it should need only 2 if we have the optimized setting for `ndots` which is 1. To be able to parameterize this configuration we need the field this PR adds. I can provide more details if you are interested.

Thanks for the context Stavros, makes sense :+1:

phoinixgrr:
🎖️  Works fine. Great work 👍 

Non blocking recommendation following: 
If we interested in automatically generating README file, documenting the chart values, their defaults, and an optional description parsed from comments, worth taking a  looking at https://github.com/norwoodj/helm-docs

Low hanging fruit 🥝 🍌 🍒

phoinixgrr (approved):

stafot:
@phoinixgrr Totally agree, let's create a ticket for this I am more than happy to work on this.

PR #358, opened 2022-07-27
Migrate CircleCI PR pipelines to Github Actions.

pfltdv:
#### Summary
We will move CircleCI pipelines to Github Actions to fix race condition. Since we stop supporting CircleCI, we are migrating existing pipelines to Github Actions.

This PR contains migration of pull request pipelines, after having stable pipelines here, we will migrate release pipelines and fix race condition at followup PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-3709

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,40 +1,51 @@
-.PHONY: install clean package ltvalues
+DOCKER := $(shell which docker)
+DOCKER_OPTS += --rm -u $$(id -u):$$(id -g)
+CLUSTER_NAME ?= mattermost-helm-test
 
-DIST_ROOT=dist
-KUBE_INSTALL := $(shell command -v kubectl 2> /dev/null)
-HELM_INSTALL := $(shell command -v helm 2> /dev/null)
+DOCKER_IMAGE_SHELLCHECK ?= koalaman/shellcheck:v0.8.0@sha256:4c4427336d2b4bdb054a1e97396fa2e9ca0c329a1f43f831b99bcaae4ac24fcd
+# Official image of chart-testing[https://github.com/helm/chart-testing].
+# Contains chart-testing cli,  helm and kubectl.
+DOCKER_IMAGE_CT ?= quay.io/helmpack/chart-testing:v3.7.0@sha256:2f87e56a0cebc6d9bb78d51cb6adac97858afb0887cac4d65f7f8c6a16054568
 
-SUB_CHARTS := $(shell ls mattermost-helm/charts)
+VERIFY_SCRIPTS += /src/tests/e2e-kind.sh
 
-all: package
+CT_CONFIG ?= tests/ct.yaml
+KIND_CONFIG ?= tests/kind-config.yaml
+CLUSTER_NAME ?= mattermost-helm-test
 
-check:
-ifndef KUBE_INSTALL
-    $(error "kubectl is not available please install from https://kubernetes.io/")
-endif
-ifndef HELM_INSTALL
-    $(error "helm is not available please install from https://github.com/kubernetes/helm")
-endif
+ .PHONY: lint
+ lint: lint-scripts lint-charts ## Lint helm charts and shell scripts
 
-.init:
-	helm init --client-only
-	touch $@
+ .PHONY: lint-charts
+lint-charts: ## Validate helm charts
+	$(DOCKER) run \
+		${DOCKER_OPTS} \
+		-v $(PWD):/src \
+		-w /src \
+		--entrypoint '/bin/sh' \
+		${DOCKER_IMAGE_CT} \
+		-c \
+		"ls >/dev/null && ct lint --config tests/ct.yaml"
 
-package: check .init
-	# Temporary while we wait for changes to be merged upstream
-	helm repo add mattermost-mysqlha https://releases.mattermost.com/helm/mysqlha
-	mkdir -p dist
-	rm -f mattermost-helm/charts/*.tgz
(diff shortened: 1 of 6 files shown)


spirosoik (commented):
great work. Do you think we can move everything under makefile too. So we can keep consistency on the way we run commands in CI?

@phoinixgrr any feedback here?

pfltdv:
> great work. Do you think we can move everything under makefile too. So we can keep consistency on the way we run commands in CI?
> 
> @phoinixgrr any feedback here?

Yes we can migrate everything as makefile but for some actions we need to to extra work. IE we can't use `ludeeus/action-shellcheck@94e0aab03ca135d11a35e5bfc14e6746dc56e7e9` Github Action and install shellcheck via make. 

Using make will also remove external Github Action dependencies.

phoinixgrr:
> great work. Do you think we can move everything under makefile too. So we can keep consistency on the way we run commands in CI?
> 
> @phoinixgrr any feedback here?

Great point. 

If we move forward with the Makefile proposal, we will loose the wealth of Publicly available actions, (reusable pieces of code), that can be used to promote fast pipeline deployment. This is actually a well advertised and selling point for GitHub actions. 
 
For every case we will need to write our own Makefile Code re-invent the wheel. 

We can maintain a repository of our own GitHub actions that can be re-used.

pfltdv:
@spirosoik @phoinixgrr I've created make targets to execute those actions in local environment.

phoinixgrr (commented):
If we are moving forward with Makefile, we should decouple from using direct helm/kubectl binaries from the host, and use a container instead.

pfltdv:
> If we are moving forward with Makefile, we should decouple from using direct helm/kubectl binaries from the host, and use a container instead.

Good suggestion, I will change makefile to execute everything in containers.

pfltdv:
@phoinixgrr @spirosoik I've made changes to run everything in docker. Our requirements are  `make`, `kind` and `docker` which are mentioned at Contributors document. PR is ready for review.

Makefile
```diff
@@ -1,40 +1,46 @@
...
+
+.PHONY: chart-e2e-tests
+chart-e2e-tests: ## Execute e2e tests on changed charts
+	 DOCKER_IMAGE_CT="$(DOCKER_IMAGE_CT)" \
+	 CLUSTER_NAME="$(CLUSTER_NAME)" \
+	 CT_CONFIG="$(CT_CONFIG)" \
+	 KIND_CONFIG="$(KIND_CONFIG)" \
+	 	tests/e2e-kind.sh
+
+.PHONY: lint-charts
```
phoinixgrr, line 24: As discussed, this fails under MacOS , returning non zero exit code. 

Workaround proposal, for moving things forward: 
```
 24 .PHONY: lint-charts
 25 lint-charts: ## Lint helm charts
 26         $(DOCKER) run \
 27                 ${DOCKER_OPTS} \
 28                 --entrypoint '/bin/sh' \
 29                 -v $(PWD):/src \
 30                 -w /src \
 31                 ${DOCKER_IMAGE_CT} \
 32                 -c \
 33                 "ls >/dev/null && ct lint --config ${CT_CONFIG}"
 34
 ```

File: Makefile [hunk 2]
```diff
@@ -1,40 +1,46 @@
...
+# Official image of chart-testing[https://github.com/helm/chart-testing].
+# Contains chart-testing cli,  helm and kubectl.
+DOCKER_IMAGE_CT ?= quay.io/helmpack/chart-testing:v3.7.0@sha256:2f87e56a0cebc6d9bb78d51cb6adac97858afb0887cac4d65f7f8c6a16054568
+
+VERIFY_SCRIPTS += /src/tests/e2e-kind.sh
+
+CT_CONFIG ?= tests/ct.yaml
+KIND_CONFIG ?= tests/kind-config.yaml
+CLUSTER_NAME ?= mattermost-helm-test
+
```
phoinixgrr, line 15: A proposal for maybe decoupling these to more generic targets. 
Like `make test` `make lint` . 

To have standardized targets for Makefiles . It will be easy to also add additional tests/lints in the future.

README.md
```diff
@@ -5,10 +5,13 @@ This repository collects a set of [Helm](https://helm.sh) charts curated by [Mat
...
+- [focalboard](charts/focalboard/)
+- [mattermost-chaos-engine](charts/mattermost-chaos-engine/)
 - [mattermost-enterprise-edition](charts/mattermost-enterprise-edition/)
-- [mattermost-push-proxy](charts/mattermost-push-proxy/)
 - [mattermost-operator](charts/mattermost-operator/)
+- [mattermost-push-proxy](charts/mattermost-push-proxy/)
+- [mattermost-rtcd](charts/mattermost-rtcd/)
+- [mattermost-team-edition](charts/mattermost-team-edition/)
 
 ## Usage
```
phoinixgrr, line 16: Maybe add a section to document  locally testing/linting?

pfltdv: Testing is mentioned in contributors guide since we have a specific document for that. I will add lint section.

image

pfltdv: After second thought, better to have those in the REadme.md to improve local development experience. 

I moved those to the Readme.md as requested.

Makefile
pfltdv: Applied

phoinixgrr (approved):
Awesome work Mustafa🥇 
A new era is born :shipit:

spirosoik (approved):

PR #360, opened 2022-08-04
Migrate CircleCI release pipelines to GitHub

pfltdv:
#### Summary
We will move CircleCI pipelines to Github Actions to fix race condition. Since we stop supporting CircleCI, we are migrating existing pipelines to Github Actions.

This PR contains migration of release pipelines.

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-3709

#### Failing CircleCI checks
Please ignore CircleCI checks, there is no circle pipeline configured in PR that causes builds to fail in CircleCI. That is because project is enabled in CircleCI.

After merging this PR to master we will disable builds at CircleCI.


.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
+name: "Release"
+
+on:
+  push:
+    branches:
+      - master
+
+concurrency: release
```
pfltdv, line 8: This will solve race condition between two PR's and allow only one PR to run at the same time.

It is ok to cancel other pending pipelines if there is new commit since release script will release all changed charts and update index after.

phoinixgrr: I like this. 
:github: rocks 🎸

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  release-charts:
+    name: Release Charts
+    runs-on: ubuntu-latest
+    
+    steps:
+      # Checkout everything since ct uses git diff to determine charts
+      - name: Checkout
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b #v3.0.2
```
phoinixgrr, line 18: Non blocking. 

We should rethink this approach of manually pinning actions. 
We are loosing all the goodies that come with dependabot.

Is there a reason for pinning those along with docker images?

Makefile
```diff
@@ -35,6 +45,14 @@ lint-scripts: ## Validate shell scripts
...
 		${DOCKER_IMAGE_SHELLCHECK} \
 		$(VERIFY_SCRIPTS)
+.PHONY: release
+release: ## Release Changed Helm Charts
+	DOCKER_IMAGE_CT="$(DOCKER_IMAGE_CT)" \
+	DOCKER_IMAGE_CR="${DOCKER_IMAGE_CR}" \
+	GIT_USER="${GIT_USER}" \
+	GIT_EMAIL="${GIT_EMAIL}" \
+	 	scripts/release.sh
+
```
phoinixgrr, line 55: Would propose this part of code, to be performed in a Github Action. 
The release process, is not something that will be run in development phase. 

Only the build/release system will have to run it. 

It will also, make the release.sh script obsolete. 
No need to bugfix/maintain it.

.github/workflows/release.yml
spirosoik: just a good practice actually to not get something which can be vulnerable and still is not detected.

Makefile
```diff
@@ -7,12 +7,19 @@ DOCKER_IMAGE_SHELLCHECK ?= koalaman/shellcheck:v0.8.0@sha256:4c4427336d2b4bdb054
 # Contains chart-testing cli,  helm and kubectl.
 DOCKER_IMAGE_CT ?= quay.io/helmpack/chart-testing:v3.7.0@sha256:2f87e56a0cebc6d9bb78d51cb6adac97858afb0887cac4d65f7f8c6a16054568
 
-VERIFY_SCRIPTS += /src/tests/e2e-kind.sh
+# Official image of chart-releaser[https://github.com/helm/chart-releaser].
+# Contains chart-releaser cli.
+DOCKER_IMAGE_CR ?= quay.io/helmpack/chart-releaser:v1.4.0@sha256:f9e6b06a1d60bcd94a2a6a4988a59233936239347c69b4ab2d99916ffc848a5b
+
+VERIFY_SCRIPTS += /src/tests/e2e-kind.sh /src/scripts/release.sh
```
spirosoik, line 14: just a comment to what is it about

Makefile
spirosoik: if there is a bug in release build how we can reproduce what CI did? we need to go to actions source code and understand what command it uses right?

spirosoik (approved):
few minor comments

.github/workflows/release.yml
pfltdv: It is recommended by Github Security Best Practices.
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Makefile
pfltdv: Having that script helped alot during development to test locally. Also it will help alot to fix the future problems. If we move everything to the Github Actions, it will be very hard for us to make development.

.github/workflows/release.yml
phoinixgrr: Without auditing the source source code of the action (as described in the above documentation), I think this does not provide any actual benefit. 
But I am not strongly opinionated on this. 

It should be nice having dependabot running on our pipelines. It is actually thought of as of the best practises by security experts.

Makefile
phoinixgrr: I'd challenge that. 
https://github.com/nektos/act

There is also possibility to connected to ghaction instance for debugging reasons. 
https://github.com/marketplace/actions/debugging-with-ssh

From my point of view, less/zero code maintenance, is the better approach here. 
But not strongly opinionated.

.github/workflows/release.yml
phoinixgrr: https://github.blog/2022-04-28-best-practices-to-keep-your-projects-secure-on-github/#respond-to-vulnerable-dependencies

phoinixgrr:
I guess we need to disable CircleCI pipelines, to be able to merge this.

Makefile
phoinixgrr: @spirosoik @pfltdv  
Imaging there is a vurnerability introduced, in a binary that the script runs. 

Q: How will we know about this? 
Q: Do you think there will be enough time from detection, to actual commiting all of our repos that contain the vurenrability?
Q: No standardization, the release code commited in the makefile, can be whatever.

Makefile
pfltdv: it contains list of scripts to be verified. `lint-scripts` target will use that variable to execute shellcheck. Maybe we should change the naming.

pfltdv: Comment is added.

pfltdv:
> I guess we need to disable CircleCI pipelines, to be able to merge this.

Those checks are created automatically by CircleCI since this PR has no configuration. We can not disable circleci without merging this PR. (master still has circle-ci.yml and we need those pipelines to create releases until this pr is merged).

Will disable circleci integration after merging.

Makefile
pfltdv: @phoinixgrr Thanks alot for sharing details about act. Looks very promising tool. Is it possible to discuss to use act and makefile in next DevOps meeting and apply that decision into our Unified CI document as a policy. And change the pipeline at follow-up iteration if we decide to use that tool. Since the scope of this ticket to fix race condition and drop CircleCI pipelines.

phoinixgrr (approved):

phoinixgrr:
https://github.com/mattermost/mattermost-helm/pull/360#discussion_r940280062

@spirosoik 
Do check the proposal here: https://github.com/mattermost/mattermost-helm/pull/360#discussion_r940317891

I would strongly suggest rethinking using Makefiles binded with bash scripts committed to repos codebase, to build/release software.

Makefile
spirosoik: This looks really nice. let's add this in the next meeting and may a demo as Mustafa mentioned.

PR #368, opened 2022-09-23
Simplify release script and skip existing releases

spirosoik:
#### Summary
We simplified the script and utilising better the `chart-releaser` CLI. Now we iterate through all charts and we package them but we will upload only the packages are new. Added a new flag for `chart-releaser` which will help us to skip the existing packages.

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-3970

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -13,7 +13,7 @@ main() {
 
     for chart in charts/*; do
         echo "Packaging chart '$chart'..."
-        package_chart $chart
+        package_chart "$chart"
     done
     release_charts
     sleep 30


phoinixgrr:
Great work here. 

I'm always in font of simplifying things 👍 
Shellcheck comes green, no objections merging this

phoinixgrr (approved):

spirosoik:
@pfltdv I would like your feedback first before I merge. I removed the complexity and I just followed what everyone does out there actually for the public helm charts (deliveryhero, grafana, prometheus etc.)

scripts/release.sh
```diff
@@ -82,7 +48,7 @@ release_charts() {
         -w /src \
         -e CR_TOKEN="${CR_TOKEN}" \
         "${DOCKER_IMAGE_CR}" \
-        upload -o mattermost -r mattermost-helm
+        upload --skip-existing -o mattermost -r mattermost-helm
```
pfltdv, line 51: :heart: I love this flag. Great work Spiros.

pfltdv (approved):

pfltdv:
> @pfltdv I would like your feedback first before I merge. I removed the complexity and I just followed what everyone does out there actually for the public helm charts (deliveryhero, grafana, prometheus etc.)

Amazin work @spirosoik . It is so simple now with only adding one flag.

PR #373, opened 2022-10-12
Feature/secret

patatman:
#### Summary

This adds the ability to specify a registry secret similar to the mattermost-operator helm chart. I was having issues with specifying the registry secret, this PR makes it more clear in the helm-values.yaml. 

#### Ticket Link

Could this be included with the `hacktoberfest` label?

### DIFF
--- a/charts/mattermost-rtcd/Chart.yaml
+++ b/charts/mattermost-rtcd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: mattermost-rtcd
 description: A Helm chart for Kubernetes to deploy Mattermost's RTCD
 type: application
-version: 1.1.0
+version: 1.1.1
 appVersion: latest
 keywords:
 - mattermost
--- a/charts/mattermost-rtcd/templates/daemonset.yaml
+++ b/charts/mattermost-rtcd/templates/daemonset.yaml
@@ -23,9 +23,9 @@ spec:
       labels:
         {{- include "mattermost-rtcd.selectorLabels" . | nindent 8 }}
     spec:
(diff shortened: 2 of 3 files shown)


angeloskyratzakos (approved):
Thank you @patatman for the suggestion! LGTM

spirosoik:
hey @patatman could you please elaborate what was the issue you faced? If you keep `imagePullSecrets` empty there shouldn't be any issue.

patatman:
The issue is in the values.yaml when you do want to use a registry secret.
The current setup requires a list of registry secrets, which isn't normally the case. You'd only have one registry secret for that specific image. 

This PR makes this more clear, and uses the same setup as the other helm chart (and unifying the config with the different helm charts).

Not a breaking issue currently, but a minor improvement.

spirosoik:
> The issue is in the values.yaml when you do want to use a registry secret. The current setup requires a list of registry secrets, which isn't normally the case. You'd only have one registry secret for that specific image.
> 
> This PR makes this more clear, and uses the same setup as the other helm chart (and unifying the config with the different helm charts).
> 
> Not a breaking issue currently, but a minor improvement.

I believe this actually will limit any changes in the future. If for any reason we need to add another container like `initContainer` and we need another secret for a different registry this will limit us actually right?

patatman:
Yeah I do think that limits it, but not for the init container. Since this registry secret is used for the pod, and a container(including an init container) is tied to a pod. 

But I do think you'll have the same issue with the current setup.

patatman:
Because this limits the use of only one secret, I discussed with @spirosoik that the current implementation is more suited for future changes.
Going to cancel this PR.

PR #374, opened 2022-10-12
Fix update index function to skip if there is no release

spirosoik:
#### Summary
When we added the flag `--skip-existing` the `update_index` started failing as there are no changes in the `index.yaml` if there is no release in practice by any helm.

#### Ticket Link
Ticket: https://mattermost.atlassian.net/browse/CLD-4194

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -5,12 +5,15 @@ set -o nounset
 set -o pipefail
 
 readonly REPO_ROOT="${REPO_ROOT:-$(git rev-parse --show-toplevel)}"
+CR_INDEX_DIR=.cr-index
+CR_RELEASE_PACKAGES_DIR=.cr-release-packages
 
 main() {
     pushd "$REPO_ROOT" > /dev/null
     git config user.name "${GIT_USER}"
     git config user.email "${GIT_EMAIL}"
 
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -5,12 +5,15 @@ set -o nounset
...
 readonly REPO_ROOT="${REPO_ROOT:-$(git rev-parse --show-toplevel)}"
+CR_INDEX_DIR=.cr-index
+CR_RELEASE_PACKAGES_DIR=.cr-release-packages
 
 main() {
     pushd "$REPO_ROOT" > /dev/null
     git config user.name "${GIT_USER}"
     git config user.email "${GIT_EMAIL}"
 
+    rm -rf $CR_RELEASE_PACKAGES_DIR $CR_INDEX_DIR && mkdir -p $CR_RELEASE_PACKAGES_DIR $CR_INDEX_DIR
```
phoinixgrr, line 16: Why are we deleting these folders, and recreating them afterwards? 
If we are only interested in deleting the contents?

spirosoik: we can do this too. I just reverted back the old line actually. In any case they are not committed as we excluded them in `.gitignore`.

phoinixgrr (approved):

PR #377, opened 2022-10-18
Fix regcred

patatman:
#### Summary

This adds the ability to add multiple Private registry secrets when needed. 
The old way of adding secrets was limited to just one.


patatman:
Ping: @phoinixgrr and @spirosoik 
This makes the regcred setup the same as the `mattermost-rtcd` helm chart.

PR #380, opened 2022-11-09
Readme: Fix broken URL to values.yaml

gkech:
#### Summary

While exploring the documentation for deploying the operator, I noticed that there is a broken URL in its readme that we can fix.


phoinixgrr (approved):
Nice catch. Thanks @gkech  🎖️

stylianosrigas (approved):
Thanks LGTM :)

PR #383, opened 2022-11-16
Enables saml setting to be overridden

KhizerJaan:
#### Summary

Previously , saml settings were not configured to be overridden , it would enable saml settings to be overridden.


erezo9:
@KhizerJaan the config.tpl doesnt apply to the deployment unless you add the config into a volumemount and volume
since its a Kubernetes installation, the configuration is stored in the database and the whole point of the config file is irrelavent

if you see in the code you can see the config map is only applied to the load test and not the application it self
hope this helps

PR #385, opened 2022-12-06
add customization for rtcd deployment

erezo9:
This PR should add customization to the deployment to allow the user to choose between deployment and daemonset
also adds the possibilty to use nodeport instead of cluster ip
also removes the use of ingress since udp data cant go through layer 7 ingress


charts/mattermost-rtcd/templates/deploy.yaml
```diff
@@ -9,11 +9,11 @@ spec:
     matchLabels:
       {{- include "mattermost-rtcd.selectorLabels" . | nindent 6 }}
   updateStrategy:
```
InsaneTimeWatcher, line 11: need fix with replicas https://github.com/erezo9/mattermost-helm/pull/1

InsaneTimeWatcher (changes requested):
need fix updateStrategy and typo replicas

stylianosrigas:
@erezo9 This PR can be closed, as this feature is already supported in chart version 1.2.1

PR #397, opened 2023-04-20
Convert RTCD to deployments from Daemonsets

d-wierdsma:
#### Summary

Convert RTCD to deployments from Daemonsets

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-5537

### DIFF
--- a/charts/mattermost-rtcd/values.yaml
+++ b/charts/mattermost-rtcd/values.yaml
@@ -36,7 +36,7 @@ securityContext: {}
 deployment:
   replicas: 2
 
-  environmentVariables:
+  environmentVariables: {}
     # All env vars that exist for RTCD: https://github.com/mattermost/rtcd/blob/master/docs/env_config.md
     # RTCD_API_SECURITY_ALLOWSELFREGISTRATION: "\"true\""
     # RTCD_LOGGER_CONSOLELEVEL: "\"DEBUG\""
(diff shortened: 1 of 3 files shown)


charts/mattermost-rtcd/values.yaml
```diff
@@ -95,4 +97,15 @@ tolerations: {}
 
 dnsConfig: {}
 
-affinity: {}
+affinity:
```
stylianosrigas, line 100: @d-wierdsma We need an affinity here that ensures that not both pods will be scheduled in the same rtcd node as calls will not work with that.

d-wierdsma: Updated change

stafot (approved):

stafot:
You could make replicas number configurable but 0/5.

d-wierdsma:
> You could make replicas number configurable but 0/5.

@stafot I believe I did here https://github.com/mattermost/mattermost-helm/pull/397/files#diff-0e226ccb0251623c490fd79a8dfa867c8039a9a99dd6e1fba48b053a3f3020c0R37

angeloskyratzakos (approved):

streamer45:
@d-wierdsma Do any of these changes require updating our documentation in https://docs.mattermost.com/configure/calls-deployment.html#kubernetes-deployments?

d-wierdsma:
Ah shoot yes, can you link me where those docs are stored and I can update them?

d-wierdsma:
Thanks for bringing that to my attention Claudio!

Here is the PR https://github.com/mattermost/docs/pull/6350

stafot:
@d-wierdsma Since we had as default method of deploying calls `daemonSet` in my view we should support it for sometime as might other people use it. Since this PR introduces a breaking change a safer approach would be to support both types of  deploying resources (`daemonsets` and `deployments`). Initially we could leave as default `daemonset` and be able with a flag to use `deployment` and at later step do default deployment and second way `daemonset`. I apologize because I did this comment after the merge of this code.

PR #406, opened 2023-06-19
[MM-53140] Allow TCP connectivity for Calls

streamer45:
#### Summary

We are introducing support for TCP candidates for calls in https://github.com/mattermost/rtcd/pull/103. This means that if no UDP connection is possible, a TCP channel can be used instead as a backup option. The big advantage of this approach is that it can avoid having to deploy and maintain a separate TURN server to guarantee connectivity for clients unable to use UDP for whatever reason (e.g. protocol blocked by firewall).

I am updating the helm chart to reflect the changes although I couldn't really understand whether most of these are actually used/needed given we use host networking. Looking for SREs input on this.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53140

### DIFF
--- a/charts/mattermost-rtcd/templates/daemonset.yaml
+++ b/charts/mattermost-rtcd/templates/daemonset.yaml
@@ -55,10 +55,14 @@ spec:
               containerPort: {{ .Values.service.APIport }}
               hostPort: {{ .Values.service.APIport }}
               protocol: TCP              
-            - name: rtc
+            - name: rtc-udp
               containerPort: {{ .Values.service.RTCport }}
               hostPort: {{ .Values.service.RTCport }}
               protocol: UDP
+            - name: rtc-tcp
+              containerPort: {{ .Values.service.RTCport }}
+              hostPort: {{ .Values.service.RTCport }}
+              protocol: TCP
           livenessProbe:
             httpGet:
               path: /version
--- a/charts/mattermost-rtcd/values.yaml
+++ b/charts/mattermost-rtcd/values.yaml
@@ -44,8 +44,11 @@ configuration:
   environmentVariables: {}
     # All env vars that exist for RTCD: https://github.com/mattermost/rtcd/blob/master/docs/env_config.md
     # RTCD_API_SECURITY_ALLOWSELFREGISTRATION: "\"true\""
+    # RTCD_RTC_ICESERVERS: "'[{\"urls\":[\"stun:stun.global.calls.mattermost.com:3478\"]}]'"
     # RTCD_LOGGER_CONSOLELEVEL: "\"DEBUG\""
     # RTCD_LOGGER_ENABLEFILE: "\"false\""
+    # RTCD_RTC_ICEPORTUDP: "\"8443\""
+    # RTCD_RTC_ICEPORTTCP: "\"8443\""
 
   maxUnavailable: 1     # Only used when updateStrategy is set to "RollingUpdate"
   updateStrategy: RollingUpdate
@@ -64,7 +67,7 @@ configuration:
 service:
   # APIport is the port that websocket and API uses
   APIport: 8045
-  # RTCport is the UDP port that RTC uses
+  # RTCport is the port (both UDP and TCP) that will serve calls related traffic (i.e. audio,video)
   RTCport: 8443
 
 ingress:
--- a/charts/mattermost-rtcd/Chart.yaml
+++ b/charts/mattermost-rtcd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: mattermost-rtcd
 description: A Helm chart for Kubernetes to deploy Mattermost's RTCD
 type: application
-version: 1.3.0
+version: 1.4.0
 appVersion: latest
 keywords:
 - mattermost
--- a/charts/mattermost-rtcd/Chart.yaml
+++ b/charts/mattermost-rtcd/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: mattermost-rtcd
 description: A Helm chart for Kubernetes to deploy Mattermost's RTCD
 type: application
-version: 1.2.1
+version: 1.3.0
 appVersion: latest
 keywords:
 - mattermost
--- a/charts/mattermost-rtcd/templates/deployment.yaml
+++ b/charts/mattermost-rtcd/templates/deployment.yaml
@@ -56,10 +56,14 @@ spec:
               containerPort: {{ .Values.service.APIport }}
               hostPort: {{ .Values.service.APIport }}
               protocol: TCP              
-            - name: rtc
+            - name: rtc-udp
               containerPort: {{ .Values.service.RTCport }}
               hostPort: {{ .Values.service.RTCport }}
               protocol: UDP
+            - name: rtc-tcp
+              containerPort: {{ .Values.service.RTCport }}
+              hostPort: {{ .Values.service.RTCport }}
+              protocol: TCP
           livenessProbe:
             httpGet:
               path: /version
--- a/charts/mattermost-rtcd/templates/service.yaml
+++ b/charts/mattermost-rtcd/templates/service.yaml
@@ -12,8 +12,12 @@ spec:
       protocol: TCP
       name: api
     - port: {{ .Values.service.RTCport }}
-      targetPort: rtc
+      targetPort: rtc-udp
       protocol: UDP
-      name: rtc
+      name: rtc-udp
+    - port: {{ .Values.service.RTCport }}
+      targetPort: rtc-tcp
+      protocol: TCP
+      name: rtc-tcp
   selector:
     {{- include "mattermost-rtcd.selectorLabels" . | nindent 4 }}


charts/mattermost-rtcd/values.yaml
```diff
@@ -44,8 +44,11 @@ configuration:
   environmentVariables: {}
     # All env vars that exist for RTCD: https://github.com/mattermost/rtcd/blob/master/docs/env_config.md
     # RTCD_API_SECURITY_ALLOWSELFREGISTRATION: "\"true\""
+    # RTCD_RTC_ICESERVERS: "'[{\"urls\":[\"stun:stun.global.calls.mattermost.com:3478\"]}]'"
     # RTCD_LOGGER_CONSOLELEVEL: "\"DEBUG\""
     # RTCD_LOGGER_ENABLEFILE: "\"false\""
+    # RTCD_RTC_ICEPORTUDP: "\"8443\""
+    # RTCD_RTC_ICEPORTTCP: "\"8443\""
```
streamer45, line 51: Adding these because no matter what you configure as `RTCport` below these are the actual ports the daemon will listen on and should be updated accordingly if needed.

streamer45:
Another slightly unrelated question but I see we include the ingress template in the chart files but then explicitly disable it. It sounds like that's legacy stuff that we won't need going forward as we don't really support running Calls through an ingress. I wonder if we should consider removing it to avoid the extra confusion.

cpoile (approved):

charts/mattermost-rtcd/templates/daemonset.yaml
```diff
@@ -55,10 +55,14 @@ spec:
               containerPort: {{ .Values.service.APIport }}
               hostPort: {{ .Values.service.APIport }}
               protocol: TCP              
-            - name: rtc
+            - name: rtc-udp
               containerPort: {{ .Values.service.RTCport }}
```
spirosoik, line 59: this won't work as we use the same port and we have the `hostPort`. The ports will overlap each other.

streamer45: I tested a bit locally and could get it to expose it this way. The only caveat was that minikube doesn't support mixed protocols on anything but port `443` so I used that one (https://minikube.sigs.k8s.io/docs/tutorials/nginx_tcp_udp_ingress/#caveats).

```
Name:         mattermost-rtcd-n8gx5
Namespace:    default
Priority:     0
Node:         minikube/192.168.49.2
Start Time:   Wed, 21 Jun 2023 10:41:53 -0600
Labels:       app.kubernetes.io/instance=mattermost-rtcd
              app.kubernetes.io/name=mattermost-rtcd
              controller-revision-hash=5bc89cc7f5
              pod-template-generation=1
Annotations:  <none>
Status:       Running
IP:           192.168.49.2
IPs:
  IP:           192.168.49.2
Controlled By:  DaemonSet/mattermost-rtcd
Containers:
  mattermost-rtcd:
    Container ID:   docker://74cfd07566702e8693c120a8ca9b9bba74700235698ad7abdac1942ebac9a245
    Image:          mattermost/rtcd-daily:dev-abfdcea
    Image ID:       docker-pullable://mattermost/rtcd-daily@sha256:84dc49cfe6fc5e578b37ca8b7338456ec59fe6fb77a297cc42c1f0b531da5480
    Ports:          8045/TCP, 443/UDP, 443/TCP
    Host Ports:     8045/TCP, 443/UDP, 443/TCP
    State:          Running
      Started:      Wed, 21 Jun 2023 10:41:53 -0600
    Ready:          True
    Restart Count:  0
    Limits:
      cpu:     1
      memory:  2Gi
    Requests:
      cpu:      100m
      memory:   32Mi
    Liveness:   http-get http://:api/version delay=0s timeout=1s period=10s #success=1 #failure=3
    Readiness:  http-get http://:api/version delay=0s timeout=1s period=10s #success=1 #failure=3
    Environment:
      RTCD_API_SECURITY_ALLOWSELFREGISTRATION:  true
      RTCD_LOGGER_CONSOLELEVEL:                 DEBUG
      RTCD_LOGGER_ENABLEFILE:                   false
      RTCD_RTC_ICEPORTTCP:                      443
      RTCD_RTC_ICEPORTUDP:                      443
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-d69tz (ro)
Conditions:
  Type              Status
  Initialized       True 
  Ready             True 
  ContainersReady   True 
  PodScheduled      True 
Volumes:
  kube-api-access-d69tz:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           kube-root-ca.crt
    ConfigMapOptional:       <nil>
    DownwardAPI:             true
QoS Class:                   Burstable
Node-Selectors:              <none>
Tolerations:                 node.kubernetes.io/disk-pressure:NoSchedule op=Exists
                             node.kubernetes.io/memory-pressure:NoSchedule op=Exists
                             node.kubernetes.io/network-unavailable:NoSchedule op=Exists
                             node.kubernetes.io/not-ready:NoExecute op=Exists
                             node.kubernetes.io/pid-pressure:NoSchedule op=Exists
                             node.kubernetes.io/unreachable:NoExecute op=Exists
                             node.kubernetes.io/unschedulable:NoSchedule op=Exists
                             rtcd=true:NoSchedule
Events:
  Type    Reason     Age   From               Message
  ----    ------     ----  ----               -------
  Normal  Scheduled  8s    default-scheduler  Successfully assigned default/mattermost-rtcd-n8gx5 to minikube
  Normal  Pulled     9s    kubelet            Container image "mattermost/rtcd-daily:dev-abfdcea" already present on machine
  Normal  Created    9s    kubelet            Created container mattermost-rtcd
  Normal  Started    9s    kubelet            Started container mattermost-rtcd
```

But more importantly, do we actually need to expose the ports there when `hostNetwork` is `true`? My very basic understanding is that we don't but would like to hear from you guys.

streamer45: @spirosoik @stylianosrigas Not urgent but would be nice to sort this out whenever you get a chance so that eventually we could get it working on Community as we had a few people unable to join in the past due to client-side UDP restrictions.

Also happy to jump on a call if necessary, just let me know :)

spirosoik:
@streamer45 thanks for letting us know, I think we need to involve @iyampaul as part of our change management process.

Is there any updated design doc?

streamer45:
> @streamer45 thanks for letting us know, I think we need to involve @iyampaul as part of our change management process.
> 
> Is there any updated design doc?

@spirosoik I don't think there's much to change. The [original design doc](https://docs.google.com/document/d/1XRUCTb-sCp2573VhbaOZST6oxJwd25ANQ1IT1WYT5sE/edit) already included the option for TCP connectivity but we never added support for it, until now. I think we mostly need to update some documentation and deployment files like this one (docker ones were updated already in https://github.com/mattermost/docker/pull/131).

There isn't anything else changing. This is purely about giving clients an option to connect through a pre-configured port using TCP when UDP is not available for whatever reason.

Of course to deploy this in Community we should engage with security because we'll likely have to expose the additional port (or protocol at least if using the same port number).

iyampaul:
@streamer45 @spirosoik - Is the intent for this to continue routing through cloudflare, or is this going direct to the origin/endpoint?  If the latter, we'll want to look at enabling AWS WAF in front of it.

> > @streamer45 thanks for letting us know, I think we need to involve @iyampaul as part of our change management process.
> > Is there any updated design doc?
> 
> @spirosoik I don't think there's much to change. The [original design doc](https://docs.google.com/document/d/1XRUCTb-sCp2573VhbaOZST6oxJwd25ANQ1IT1WYT5sE/edit) already included the option for TCP connectivity but we never added support for it, until now. I think we mostly need to update some documentation and deployment files like this one (docker ones were updated already in [mattermost/docker#131](https://github.com/mattermost/docker/pull/131)).
> 
> There isn't anything else changing. This is purely about giving clients an option to connect through a pre-configured port using TCP when UDP is not available for whatever reason.
> 
> Of course to deploy this in Community we should engage with security because we'll likely have to expose the additional port (or protocol at least if using the same port number).

streamer45:
> @streamer45 @spirosoik - Is the intent for this to continue routing through cloudflare, or is this going direct to the origin/endpoint? If the latter, we'll want to look at enabling AWS WAF in front of it.
> 
> > > @streamer45 thanks for letting us know, I think we need to involve @iyampaul as part of our change management process.
> > > Is there any updated design doc?
> > 
> > 
> > @spirosoik I don't think there's much to change. The [original design doc](https://docs.google.com/document/d/1XRUCTb-sCp2573VhbaOZST6oxJwd25ANQ1IT1WYT5sE/edit) already included the option for TCP connectivity but we never added support for it, until now. I think we mostly need to update some documentation and deployment files like this one (docker ones were updated already in [mattermost/docker#131](https://github.com/mattermost/docker/pull/131)).
> > There isn't anything else changing. This is purely about giving clients an option to connect through a pre-configured port using TCP when UDP is not available for whatever reason.
> > Of course to deploy this in Community we should engage with security because we'll likely have to expose the additional port (or protocol at least if using the same port number).

@iyampaul As far as I understand Calls traffic is **not** going through Cloudflare as I am not even sure if they support UDP. But I am not in a great position to help here as I don't have much visibility on the actual deployment specifics. So we may indeed need to add a layer on top if TCP is deemed more risky, although I am not entirely sure why that would be the case.

spirosoik:
@iyampaul they are not going through cloudflare.

spirosoik (approved):
Thanks 🚀

stylianosrigas (approved):

PR #417, opened 2023-09-04
Remove rand suffix in test job name

mustdiechik:
Also added possibility to add additional
annotations for test job.

---
It is bad practice to use random prefix\suffix in name of resources.
- helm template is always different, it is pain for linters
- it is convenient to search resource by name when use kustomize.

Also added possibility to add additional annotations, for example it is useful when use ArgoCD.


spirosoik (approved):
Thanks for the contribution

charts/mattermost-team-edition/values.yaml
```diff
@@ -247,3 +247,8 @@ readinessProbe:
   failureThreshold: 3
   timeoutSeconds: 5
   periodSeconds: 10
+
+## Additional annotations for test job
+
+testJob:
```
spirosoik, line 253: I am wondering if we need to this as an extra option. Why not just use the same value we already have `extraPodAnnotations `

mustdiechik: yes, we need. Because to manage job with ArgoCD for example, we need its own job annotations. Helm`s hook annotations would not work.
https://argo-cd.readthedocs.io/en/stable/user-guide/resource_hooks/

Also if we need to redeploy job for each "apply", we may to add some annotation with {{ rand }} value.

spirosoik: Yeah I am aware for the argo resource hooks etc. Though we already have `extraPodAnnotations` in `values.yaml` what if we just use this in the test helm template so we have one value vs creating test only values.

spirosoik: In practice do this:

```
In practice do this `{{- with .Values.extraPodAnnotations}}`
```

mustdiechik: I think, nobody will not want to add hook`s annotations that necessary for JOB to POD.

spirosoik: Could you elaborate a bit more? I think having one value for annotations and just re-use will help us to be DRY.

mustdiechik: The job should be run after POD was created or updated. When to run job I can managed via it`s annotation.

mustdiechik: @spirosoik up

spirosoik:
@stylianosrigas could you please have a look too here?

spirosoik (approved):

stylianosrigas (approved):
@mustdiechik LGTM thanks a lot for your contribution 🚀

PR #438, opened 2024-01-30
[CLD-7062] Push proxy chart revisit ingress spec

phoinixgrr:
#### Summary

This covers the case when special ingress spec is needed to manage different http paths.
For example, provides support for the following configuration:

```
          - backend:
              service:
                name: response-403
                port:
                  name: use-annotation
            path: /metrics
            pathType: Prefix
```
In which along with the following Ingress annotation and the use of AWS ALB Controller:
```
    alb.ingress.kubernetes.io/actions.response-403: >
          {"type":"fixed-response","fixedResponseConfig":{"contentType":"text/plain","statusCode":"403","messageBody":"403 Not Allowed"}}
```
Will result in different handling of specific paths

The template continues to render as usual in case `customService` directive is not in use. 

#### Ticket Link

Ticket: https://mattermost.atlassian.net/browse/CLD-7062

### DIFF
--- a/charts/mattermost-push-proxy/Chart.yaml
+++ b/charts/mattermost-push-proxy/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 description: Mattermost Push Proxy server
 name: mattermost-push-proxy
 type: application
-version: 0.10.3
+version: 0.10.4
 appVersion: 5.28.0
 keywords:
 - mattermost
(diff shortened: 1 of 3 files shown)


saturninoabril (approved):
Thanks @phoinixgrr!

toninis (approved):

stafot (approved):

mvitale1989 (approved):

PR #446, opened 2024-02-29
feat: Upgrade push proxy version and use default value from Chart.yaml

toninis:
#### Summary

- Upgrade mattermost push proxy chart to support arm64/amd64 docker images 
- Use default .Chart.AppVersion value for image tag 

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-7247

### DIFF
--- a/charts/mattermost-push-proxy/values.yaml
+++ b/charts/mattermost-push-proxy/values.yaml
@@ -1,10 +1,10 @@
 replicaCount: 1
 image:
   repository: mattermost/mattermost-push-proxy
-  tag: 5.30.0
   pullPolicy: Always
 
-resources: {}
+resources:
+  {}
   # We usually recommend not to specify default resources and to leave this as a conscious
   # choice for the user. This also increases chances charts run on environments with little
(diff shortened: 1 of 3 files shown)


charts/mattermost-push-proxy/values.yaml
```diff
@@ -1,10 +1,10 @@
 replicaCount: 1
 image:
   repository: mattermost/mattermost-push-proxy
-  tag: 5.30.0
```
spirosoik, line 4: could you elaborate why did we remove the tag?

spirosoik: ignore, you used the app version.

toninis: Because I added the default value from `appVersion` in the `Chart.yaml` . We do not have to change the version in both places

toninis:
```yaml
          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```

stafot (approved):
thanks. LGTM

PR #448, opened 2024-03-07
Allow probe scheme to be customised for TLS deployments

fiddyspence:
… HTTP/S termination options for liveness/readiness/startup probes

#### Summary
Adds a configuration option and settings in the deployment to facilitate TLS enabled pods

#### Description
Adds a ```probeScheme``` setting in values.yaml which allows the various probes to be either HTTP or HTTPS

Defaults to plain HTTP

#### Ticket Link
None

### DIFF
--- a/charts/mattermost-team-edition/templates/deployment.yaml
+++ b/charts/mattermost-team-edition/templates/deployment.yaml
@@ -87,14 +87,17 @@ spec:
           httpGet:
             path: /healthz
             port: http
+            scheme: {{ .Values.probeScheme }}
         livenessProbe: {{- .Values.livenessProbe | toYaml | nindent 10 }}
           httpGet:
             path: /api/v4/system/ping
             port: http
+            scheme: {{ .Values.probeScheme }}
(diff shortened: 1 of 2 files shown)


legoguy1000:
we need this to migrate to the helm chart

Ansible-man:
Given stuff like this is not being reviewed or merged I don't see a chance of my company adopting this application. Either poor management or lack of motivation to maintain the free version of their application. Why would we pay for something when the application cant even be maintained with the code sitting here for free.

lknite:
I would also like this as I used cloudnative-pg to stand up my postgres server and it by default creates a 'db connection string' to use which fails with mattermost because mattermost was expecting no tls.  For my use case, no tls is ok but it would be nice if it were documented to save a few hours of troubleshooting, or at least a log error which really stands out, the existing "unable to connect to db" is helpful, "tls detected when no tls was expected" would be awesome (or better yet a reference to documentation where to enable/disable tls) once this PR (or similar) is merged.

PR #457, opened 2024-05-17
[CLD-7443] Allow `rtcd` service to be exposed on a privileged port

streamer45:
#### Summary

PR proposes a possible solution to the issue we are facing when trying to have `rtcd` listen on a privileged port (lower than 1024). 

So far we've tested a couple of workarounds (full context at https://hub.mattermost.com/private-core/pl/5ubornsmpfnkuyxnmyixgfpich):

1. Setting the `NET_BIND_SERVICE` capability in the container security context.
2. Setting the `net.ipv4.ip_unprivileged_port_start` sysctl in the pod security context.

Unfortunately, none of the above is working, possibly because the image was meant to be run using an unprivileged user at all times. 

So I was thinking we could avoid the problem altogether by using the k8s service as designed and let it forward the traffic from privileged (node level) to unprivileged (container level). The only caveat is that to do so we'll need to disable host networking.

This is the bit I will need some guidance on because I don't have a full understanding of other potential side effects of doing so (e.g. DNS policy?).

> **Note**
> To make this work properly in our deployment we'll also need some changes coming in v0.15.1 (https://github.com/mattermost/rtcd/pull/140).

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-7443

### DIFF
--- a/charts/mattermost-rtcd/templates/deployment.yaml
+++ b/charts/mattermost-rtcd/templates/deployment.yaml
@@ -30,7 +30,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       dnsPolicy: ClusterFirstWithHostNet
-      hostNetwork: true
+      hostNetwork: false
       serviceAccountName: {{ include "mattermost-rtcd.serviceAccountName" . }}
       terminationGracePeriodSeconds: {{ .Values.configuration.terminationGracePeriod }}
       securityContext:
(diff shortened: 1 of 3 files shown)


charts/mattermost-rtcd/templates/deployment.yaml
```diff
@@ -30,7 +30,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       dnsPolicy: ClusterFirstWithHostNet
-      hostNetwork: true
+      hostNetwork: false
```
streamer45, line 33: If we decide to do this, I would make this a variable that can be controlled from the values file since there are still cases where it would be valuable to use host networking since it simplifies the generation of local candidates.

streamer45:
> @streamer45 Is this still something we consider to merge?

@stafot Ideally yes. We still have some clients failing to join calls on Community/Hub occasionally, likely because 8443 is blocked on their side. This could help.

stafot:
@streamer45 If we want to maerge let's do variable hostNetwork and have true as default value.

streamer45:
@stafot Sounds good. I'd also like to try this in our test environment before merging since I haven't yet validated this in a real deployment.

PR #466, opened 2024-08-07
Refactor secrets in push-proxy

stafot:
#### Summary
Refactor secrets in push-proxy

#### Ticket Link
https://mattermost.atlassian.net/browse/CLD-8017

### DIFF
--- a/charts/mattermost-push-proxy/templates/deployment.yaml
+++ b/charts/mattermost-push-proxy/templates/deployment.yaml
@@ -34,75 +34,150 @@ spec:
         - name: {{ include "mattermost-push-proxy.name" . }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          args: ["--config", "/mattermost-push-proxy/config/config.json"]
+          args:
+            - "--config"
+            - "/mattermost-push-proxy/config/config.json"
           ports:
             - containerPort: {{ .Values.service.internalPort }}
           volumeMounts:
-            - mountPath: /mattermost-push-proxy/config/config.json
-              name: push-config-template
-              subPath: push-config.json
-            # Volume Mounts handling based on externalSecrets.enabled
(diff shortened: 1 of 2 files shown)


angeloskyratzakos (approved):

charts/mattermost-push-proxy/templates/deployment.yaml
```diff
@@ -41,104 +41,68 @@ spec:
...
-            - key: push-config.json
-              path: push-config.json
-        {{- if .Values.applePushSettings.authKey }}
+              - key: push-config.json
+                path: push-config.json
+        # Secret definitions based on externalSecrets.enabled
         - name: apple-push-auth-key
           secret:
-            secretName: {{ include "mattermost-push-proxy.fullname" . }}-apple-certs
+            secretName: {{ if .Values.externalSecrets.enabled }}{{ include "mattermost-push-proxy.fullname" . }}-secret{{ else }}{{ include "mattermost-push-proxy.fullname" . }}-apple-certs{{ end }}
```
stylianosrigas, line 75: With the  removal of {{- if .Values.applePushSettings.authKey }} the secret will be create even if the values are empty, which we don't want right?

stafot: @stylianosrigas This change has the purpose when we have external secrets in place all key values to be under one AWS secrets for cost/simplicity reasons, at the same time for backwards compatibility we need to keep the previous structure exactly as is when we have not enable external secrets. That is the purpose of this if clause.

stylianosrigas: I am referring to the removal of  {{- if .Values.applePushSettings.authKey }}. This is now replaced by a single else which means that if external secrets are disabled it will create the old type secrets even if the applePushSettings.authKey is empty. Previously using an empty applePushSettings.authKey  would not generate any secret.

stafot: thanks for the catch.

stylianosrigas (approved):

PR #481, opened 2024-11-14
Allow you to set pod labels

bjorns163:
The current deployment lacks the ability to set pod labels. While this may not directly affect the usage of Mattermost, it is a common feature needed in bigger environments. This adds the ability to set them


stafot (approved):

techhazard:
@stafot Thanks for the review. Do you know in which release this will be available? It seems to be blocked by some checks.

Issue #315, opened 2022-03-12
[mattermost-team-edition] Respect custom securityContext


theAkito:
See the initial issue:

https://forum.mattermost.com/t/permission-error-but-permissions-are-permissive/13080

> Apparently, the user `mattermost` inside the Docker Image is hard-coded to user and group ID `2000`. Even though, the `mattermost` process is clearly running as the user (ID) I desire, it still uses the `mattermost` user with the ID `2000` somewhere in-between.
> 
> This is clearly a bug, as it would mean, that this app can collide with any other map, which uses the same ID for its user.

It's necessary for the Chart setup to respect the provided `securityContext`. If I provide the user and group ID `123` and the `mattermost` process actually runs with those (it actually does, I checked), then the classic `mattermost` user with the ID `2000` shouldn't remain the user in charge.

Related Issues:
#128
 #142

lucasscheepers:
Is this already solved?

theAkito:
@lucasscheepers 

[No, it's not officially fixed, yet.](https://github.com/mattermost/mattermost-server/blob/5bd223c836a71c1148a3b5990fde6f22f2fdc1ed/build/Dockerfile)

However, if you need help setting up a workaround, by using a custom image, then I can help you with that. 🙂

lucasscheepers:
> @lucasscheepers
> 
> [No, it's not officially fixed, yet.](https://github.com/mattermost/mattermost-server/blob/5bd223c836a71c1148a3b5990fde6f22f2fdc1ed/build/Dockerfile)
> 
> However, if you need help setting up a workaround, by using a custom image, then I can help you with that. 🙂

Yeah I'm very interested in your workaround! So you set up your security context while building the image I suppose?

lucasscheepers:
@theAkito Do you have time to share this with me?

theAkito:
@lucasscheepers 

I'm truly sorry for answering so late. At first I forgot about this, then I didn't have enough time. 😞 

If you still need help, you could for example send me a message at `a   -k i-- t-  o.k - i- - - -t- - - s- - -u- ---- n- -- - -----e -- - - -- - -(at)  p r o  --- t o-- n-  .  ---m     e---` (without the dashes and spaces, replacing `(at)` with `@`.

Once I understand your use case, I will probably just create a Github Gist with the solution for your setup.



> I've to add that I'm using the Mattermost Operator Helm chart including the manifest file to deploy Mattermost.

[...]

PR #482, opened 2024-11-15
fix security context at container level in hardened k8s cluster

jm-bertheas-datailor:
#### Summary
When we try to deploy [mattermost-team-edition] with helm charts into a hardened k8s cluster, it's not possible to define a custom securityContext at the container level (event if it's possible at pod level but it's not enough to start the mattermost-team-edition pod).
We have to add some custom securityContext definition on the deployment at the container level.

The fix done here, is supposed to be backward compatible.

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-helm/issues/315

### DIFF
--- a/charts/mattermost-team-edition/templates/deployment.yaml
+++ b/charts/mattermost-team-edition/templates/deployment.yaml
@@ -41,9 +41,9 @@ spec:
       tolerations:
       {{- toYaml .Values.tolerations | nindent 6 }}
       {{- end }}
-      {{- if .Values.securityContext }}
+      {{- if .Values.podSecurityContext }}
       securityContext:
-        {{- toYaml .Values.securityContext | nindent 8 }}
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
       {{- end }}
       {{- if .Values.serviceAccount.create }}
       serviceAccountName: {{ .Values.serviceAccount.name }}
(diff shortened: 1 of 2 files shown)


Ansible-man:
Is this going to get merged?

stafot:
@jm-bertheas-datailor 
Thanks for the PR — this looks great overall and I really appreciate the effort to align the naming with common conventions.

I do prefer `podSecurityContext` as it's more explicit and matches standard Kubernetes terminology. That said, I have a small concern about backward compatibility — since `securityContext` was previously used for the pod-level context, some users might be relying on that key in their existing `values.yaml` overrides.

To maintain compatibility, would it make sense to keep `securityContext` as-is for pod-level settings and introduce something like `containerSecurityContext` or `defaultContainerSecurityContext`  for the new container-level one?

Curious to hear your thoughts — I'm very much in favor of this PR and happy to approve it once we align on this detail.

cc @timstott

timstott:
Thanks for following up @stafot. The `podSecurityContext` and container `securityContext` share common properties but also have exclusive ones. The changes currrently proposed would be a breaking change.

I see two options available to us:

1. Introduce the breaking change. We can improve the upgrade experience [by failing](https://masterminds.github.io/sprig/flow_control.html) when an incompatible property exists on `securityContext`. 
2. Or as proposed add the `containerSecurityContext` setting.

I am happy to contribute to either solutions. The decision may come down to the mattermost's team policy.

gabrieljackson:
Great conversation here. After thinking about this a bit, I am also in favor of modifying the implementation so that we don't possibly introduce a breaking change for existing users. If we had to build a larger workaround to support them I would would say to proceed with this change, but all we need to do here is select a slightly different name for the container security context so that gets my vote.

stafot (changes requested):
Let's refactor this PR based on proposal nr.2

toninis:
@timstott @jm-bertheas-datailor my two cents here . 

We can keep the existing functionality with securityContext and also add it on the container level (it won't make a difference). 
Create the new `podSecurityContext` which will precede `securityContext` if it exists

timstott:
@gabrieljackson @stafot Thank you for your support. I have updated the implementation in https://github.com/mattermost/mattermost-helm/pull/498

PR #484, opened 2024-12-18
Update calls-offloader chart

streamer45:
#### Summary

Updating the `calls-offloader` chart to:

- Include the necessary environment variable needed by jobs to use persistent volumes. 
- Remove the volume mounts for the offloader service itself since it doesn't need to persist data.
- Use the [latest v0.9.0 image](https://hub.docker.com/layers/mattermost/calls-offloader/v0.9.0/images/sha256-1284ea8a453d3e447df53f304857803c20033f3f422659a6e902d320d00d5dbf)


stafot (approved):
LGTM

Issue #168, opened 2020-07-10
No Documentation on Using S3


jseiser:
There is nothing indicating this chart can be used directly with S3.  Ive merged in changes to be able to provide the service account the ability to handle IAM Access, but Im still not able to find enough documentation to actually configure this to store its data on an S3 bucket.

Our existing MM installation runs on EC2, with the /data folder on S3 and DB on RDS, we want to migrate to MM on Kubernetes and just keep using what we already have.

erezo9:
@jseiser 
don't know if matters but had the same issue
what i did is did export command on my current installation, that gave me all the needed environment variables to give in the extra environment variables in the chart

ttr:
@jseiser It should work if you keep MinIO disabled in chart.
It should load your config from DB, and verify if your MM settings for file storage are your own S3.

IAM would need to be done externally, but MM needs only standards aws s3 access keys.

rrey:
Am I wrong or the chart is missing a way to specify a custom service account that could be set on the deployment to provide the credentials for s3 ? (on AWS for instance) ?

erezo9:
> Am I wrong or the chart is missing a way to specify a custom service account that could be set on the deployment to provide the credentials for s3 ? (on AWS for instance) ?
You can specify a custom service account, but that wont help cause mattermost will not read from that
> 
> Does mm support using the pod credentials for the s3 access ?
As far as I know, the only way is via environment variables

the only way is to add an environment variable and probably add via an additional volume mount which is provided via a secret
you can read that here
https://docs.mattermost.com/administration/config-settings.html

rrey:
Thanks for the feedback. I want to use a service account to avoid providing any credential from environment and benefit from the AWS role mechanism.
I found [this issue](https://github.com/mattermost/mattermost-server/issues/13953) that talks about that. I'll continue the discussion there.

cpanato:
@jseiser are you willing to propose a PR to add this?

the-veloper:
I think the chart might lack support for using S3 in favor of using minio, due to the fact that the url is hard coded to the minio service. Correct me, if i am mistaken.

ref: https://github.com/mattermost/mattermost-helm/blob/master/charts/mattermost-enterprise-edition/templates/config.tpl#L143

PR #486, opened 2025-01-14
Add support for AWS S3

the-veloper:
#### Summary
This PR exposes the s3url and region variables of the config to the values of the helm chart.

#### Ticket Link
https://github.com/mattermost/mattermost-helm/issues/168


the-veloper:
this pr actually doesnt make sense. config.tpl actually refers to the load testing config. MM_CONFIG (if set to db url) is used to load the config from db entries

PR #503, opened 2025-05-20
Fix push-proxy custom ingress labels

stafot:
#### Summary
Fix push-proxy custom ingress labels


toninis (approved):
Sorry I missed that :/

stafot:
Lint was ok in both cases so easy to miss @toninis

